PR #60, opened 2022-12-14
New dev app using React and supporting hot refresh

rokotyan:
@reb-dev Here's the initial setup for our new dev app. 

Screen Recording 2022-12-13 at 4 22 16 PM
* It uses React and supports hot refresh, which is the thing I wanted to have for a long time, because it make the dev experience much smoother. 
* It also has search. I want any new feature (no matter how small it is) or a bug fix to have a corresponding example which we can use for automated testing in the future. Since there will be a lot of pages, navigation across them becomes important.

Can you please take a look at the code, run the app and share your thoughts?


reb-dev (commented):
@rokotyan Looks good! I really like the search bar functionality and it seems like a good setup overall.

One thing I'm not sure about is the `examples.ts` file.
1. Isn't there some way to autogenerate the routes based on the file structure? If we could eliminate one extra step of having to update this file every time we have a new example, I feel like it would be a better dev experience.
2. If not, I think the structure would look cleaner if we put this file in the `examples` directory and name it `index.ts`

Thoughts?

rokotyan:
@reb-dev Yes, that's possible with Webpack! I was going to do it later but why wait. Check out the last few commits with that change.

reb-dev (approved):
@rokotyan Nice! Looks great.

PR #61, opened 2022-12-14
Docs: Bullet Legend

reb-dev:
#63 
screencapture-localhost-9300-docs-auxiliary-BulletLegend-2022-12-14-06_44_16


rokotyan:
@reb-dev Great job, I love the level of detail!

I've pushed some minor fixes. Also the TypeScript code for the interactive legend section doesn't look right. It mentions `itemsCopy`, however there's no such variable (and I don't think it's needed here for TypeScript). I'll let you to fix it and we can merge the PR after that.

image

reb-dev:
@rokotyan good catch! I fixed it.

rokotyan (approved):

PR #72, opened 2022-12-16
Component | Timeline: Disallow negative values for line segment lengths

reb-dev:


packages/ts/src/components/timeline/index.ts
```diff
@@ -232,7 +232,7 @@ export class Timeline<Datum> extends XYComponentCore<Datum, TimelineConfig<Datum
         .attr('y', yStart + y + (config.rowHeight - height) / 2)
         .attr('width', config.showEmptySegments
           ? Math.max(config.lineCap ? height : 1, width)
-          : width)
+          : Math.max(0, width))
```
rokotyan, line 235: @reb-dev I'm thinking that if the width is negative most likely it's due to misconfiguration of the component. So it's worth adding a console warning to give users a hint. 

We usually use the `Unovis | <Component>:` prefix for warning messages:
```ts
console.warn('Unovis | Timeline: ...')
```

reb-dev: @rokotyan yes, you're right! I just added that.

rokotyan (approved):

Issue #76, opened 2022-12-19
Component | TopoJSONMap: Console errors when the container has zero width / height


rokotyan:
<img width="231" alt="image" src="https://user-images.githubusercontent.com/755708/208489310-e5a08394-09e3-4abc-940f-aad6757ef400.png">

PR #78, opened 2022-12-20
Component | TopoJSONMap: Preventing `scale` from being `NaN`

rokotyan:
Fixes console errors from #76

Component | TopoJSONMap: Preventing `scale` from being `NaN` when the container has zero width / height (e.g. when the corresponding DOM node gets hidden, triggering a ResizeObserver).


reb-dev (approved):
Looks good!

PR #90, opened 2023-01-03
Component | Graph: Fix nodeFill ignoring non-hex color values

reb-dev:
@rokotyan 

My changes in this this [commit](https://github.com/f5/unovis/commit/9dbbb56d968b7c6f7b8a1849f5d5dbdd9d03122c) produced a bug where some color values for `nodeFill` weren't recognized, resulting in icons being barely visible or not visible at all for bright node colors.

For example, providing the values `#eff` and `rgb(255,255,255)` gives us icons with the color `--vis-graph-node-icon-fill-color-bright` (before):
Screen Shot 2023-01-03 at 3 12 26 PM

When we expect `--vis-graph-node-icon-fill-color-dark` (after):
Screen Shot 2023-01-03 at 3 12 56 PM

### DIFF
--- a/packages/dev/src/examples/networks-and-flows/graph/graph-custom-fill/index.tsx
+++ b/packages/dev/src/examples/networks-and-flows/graph/graph-custom-fill/index.tsx
@@ -1,6 +1,5 @@
 import React from 'react'
 import { VisSingleContainer, VisGraph } from '@unovis/react'
-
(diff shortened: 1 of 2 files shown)


rokotyan:
@reb-dev I see, thanks for fixing!

caabernathy (approved):

rokotyan (approved):

PR #99, opened 2023-01-12
Component | XYLables: Setting stroke via CSS variables

### LINKED ISSUE #98: Component  | XYLables: Changing stroke via CSS variables doesn't work
Setting `--vis-xy-label-stroke-color` won't have any effect because the stroke value is overwritten via element styles. We don't need every element to have stroke by default, so we can simply remove that code.

rokotyan:
Fixes #98. Stroke can now be set via CSS variables:
```css
.stackBarLabels g[alert] {
  --vis-xy-label-stroke-color: white;
  --vis-xy-label-stroke-width: 1px;
}

.stackBarLabels g[alert][cluster] {
  --vis-xy-label-stroke-width: 2px;
}

```
Screen Shot 2023-01-12 at 9 49 54 AM


reb-dev (commented):
@rokotyan Should we also include CSS variables for dark theme?

rokotyan:
@reb-dev Good point, I was also thinking about it. But I'm planning some additional work for this component soon, so we can cover it then.

reb-dev (approved):
:thumbsup:

Issue #108, opened 2023-01-20
Unovis React | Components only update when `data` is passed to container


reb-dev:
### Discussed in https://github.com/f5/unovis/discussions/107

<div type='discussions-op-text'>

<sup>Originally posted by **ndtreviv** January 19, 2023</sup>
I see that you've added some [docs on selecting nodes](https://unovis.dev/docs/networks-and-flows/Graph#selecting-nodes-and-links).

How did you do the dashed border around the node on node 19 below:

<img width="272" alt="Screenshot 2023-01-19 at 10 14 07" src="https://user-images.githubusercontent.com/1530653/213415892-3ce11a17-7a49-4a56-bc0e-d249abc7ad4f.png">
</div>

rokotyan:
The fix is ready. Will be published in v1.0.3.

PR #109, opened 2023-01-20
React: Fix rendering issue

reb-dev:
#108 

Previously, our `arePropsEqual` would only check if child props had changed for multiple children, but in the case of only one child it would always return `true`. This prevented the re-rendering of the child component even though it had updated props.

Added a new dev demo page based on #107 to reflect the change.

### DIFF
--- a/packages/dev/src/examples/networks-and-flows/graph/graph-selections/index.tsx
+++ b/packages/dev/src/examples/networks-and-flows/graph/graph-selections/index.tsx
@@ -4,7 +4,7 @@ import { Graph } from '@unovis/ts'
 import { generateNodeLinkData, NodeDatum, LinkDatum } from '@src/utils/data'
 
 export const title = 'Basic Graph'
-export const subTitle = 'Generated Data'
+export const subTitle = 'Select Node on Click'
 export const category = 'Graph'
 
 export const component = (): JSX.Element => {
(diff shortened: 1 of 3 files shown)


rokotyan (approved):

rokotyan (approved):

packages/dev/src/examples/networks-and-flows/graph/graph-selections/index.tsx
```diff
@@ -4,7 +4,7 @@ import { Graph } from '@unovis/ts'
 import { generateNodeLinkData, NodeDatum, LinkDatum } from '@src/utils/data'
 
 export const title = 'Basic Graph'
-export const subTitle = 'Generated Data'
+export const subTitle = 'Select Node on Click'
```
reb-dev, line 7: @rokotyan Would you mind explaining how we should be labeling title, subTitle, and category? I want to be consistent but it's not clear to me the function of each one.

rokotyan: @reb-dev I would like them to give us a better understanding of what particular example is about. I don't have a clear guidance for that yet, so it's something we'll need to crystallize over time. If you have ideas, you're more than welcome to share them!


In the particular case:
Screen Shot 2023-01-20 at 12 04 27 PM
`category` tells us that it's a Graph example (I think this field will almost always be related to the component name);
`title` tells us that i'll be a rather basic example;
but `subTitle` give us an idea that there's something extra, i.e. node selection logic.

caabernathy (approved):

PR #116, opened 2023-01-24
Fix: Donut empty segments

reb-dev:
This PR fixes the appearance of empty segments when `padAngle` is set.

```
showEmptySegments: true
```
Screen Shot 2023-01-23 at 4 21 05 PM

```
showEmptySegments: true
padAngle: 0.02
```
Screen Shot 2023-01-23 at 4 16 58 PM

### DIFF
--- a/packages/ts/src/components/donut/index.ts
+++ b/packages/ts/src/components/donut/index.ts
@@ -15,7 +15,7 @@ import { Spacing } from 'types/spacing'
 import { VerticalAlign } from 'types/text'
 
 // Local Types
-import { DonutArcDatum, DonutArcAnimState } from './types'
+import { DonutArcDatum, DonutArcAnimState, DonutDatum } from './types'
 
 // Config
 import { DonutConfig, DonutConfigInterface } from './config'
@@ -57,38 +57,50 @@ export class Donut<Datum> extends ComponentCore<Datum[], DonutConfig<Datum>, Don
 
   _render (customDuration?: number): void {
     const { config, datamodel, bleed } = this
-    const data = datamodel.data
-    const duration = isNumber(customDuration) ? customDuration : config.duration
 
+    // Wrap data to preserve original indices
+    const data: DonutDatum<Datum>[] = datamodel.data
+      .map((d, i) => ({
+        index: i,
+        datum: d,
+      }))
+      .filter(d => config.showEmptySegments || getNumber(d.datum, config.value, d.index))
+
+    const duration = isNumber(customDuration) ? customDuration : config.duration
     const outerRadius = config.radius || Math.min(this._width - bleed.left - bleed.right, this._height - bleed.top - bleed.bottom) / 2
     const innerRadius = config.arcWidth === 0 ? 0 : clamp(outerRadius - config.arcWidth, 0, outerRadius - 1)
 
     this.arcGen
       .startAngle(d => d.startAngle)
-      .endAngle(d =>
-        // Add 0.5 degree for empty segments
-        config.showEmptySegments && d.startAngle === d.endAngle
-          ? d.startAngle + 0.5 * Math.PI / 180
-          : d.endAngle
-      )
+      .endAngle(d => d.endAngle)
       .innerRadius(d => d.innerRadius)
       .outerRadius(d => d.outerRadius)
-      .padAngle(config.padAngle)
-      .cornerRadius(d => config.cornerRadius)
+      .padAngle(d => d.padAngle)
+      .cornerRadius(config.cornerRadius)
 
-    const pieGen = pie<Datum>()
+    const pieGen = pie<DonutDatum<Datum>>()
       .startAngle(config.angleRange?.[0] ?? 0)
       .endAngle(config.angleRange?.[1] ?? 2 * Math.PI)
       .padAngle(config.padAngle)
-      .value((d, i) => getNumber(d, config.value, i) || 0)
+      .value(d => getNumber(d.datum, config.value, d.index) || 0)
       .sort(config.sortFunction)
 
     this.arcGroup.attr('transform', `translate(${this._width / 2},${this._height / 2})`)
-    const arcData = pieGen(data) as DonutArcDatum<Datum>[]
-    arcData.forEach((d, i) => {
-      d.index = i
-      d.innerRadius = innerRadius
-      d.outerRadius = outerRadius
+
+    const arcData: DonutArcDatum<Datum>[] = pieGen(data).map(d => {
+      const arc = {
+        ...d,
+        data: d.data.datum,
+        index: d.data.index,
+        innerRadius,
+        outerRadius,
+      }
+
+      if (config.showEmptySegments && d.endAngle - d.startAngle - d.padAngle <= Number.EPSILON) {
+        arc.endAngle = d.startAngle + Math.max(config.emptySegmentAngle, config.padAngle)
+        arc.padAngle = d.padAngle / 2
+      }
+      return arc
     })
 
     // Arc segments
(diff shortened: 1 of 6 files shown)


rokotyan:
@reb-dev I like your approach but it doesn't solve the wrong segment colors problem, when `showEmptySegments` is set to `false`:
image

Also, can you please combine my WIP commit with yours and update the commit history?

packages/ts/src/components/donut/index.ts
```diff
@@ -57,24 +57,19 @@ export class Donut<Datum> extends ComponentCore<Datum[], DonutConfig<Datum>, Don
 
   _render (customDuration?: number): void {
     const { config, datamodel, bleed } = this
-    const data = datamodel.data
+    const data = config.showEmptySegments ? datamodel.data : datamodel.data.filter((d, i) => getNumber(d, config.value, i))
```
rokotyan, line 60: @reb-dev I should have not committed this line, it might have confused you, sorry. That's why the colors get messed up. Let's change it back to just `data = datamodel.data`

rokotyan: @reb-dev In that case the colors will be correct, but if the donut has only one non-empty segment, where'll be a confusing gap at start:
image

reb-dev: @rokotyan Fixed! I actually like the idea of filtering the data, I tweaked it a bit so that the indices are preserved. Let me know if it's ok?

packages/ts/src/components/donut/modules/arc.ts
```diff
@@ -22,7 +22,7 @@ export function createArc<Datum> (
   config: DonutConfig<Datum>
 ): void {
   selection
-    .style('fill', (d, i) => getColor(d.data, config.color, i))
+    .style('fill', d => getColor(d.data, config.color, d.index))
```
rokotyan, line 25: @reb-dev Shouldn't it be `d.data._index` or will they be the same?

packages/ts/src/components/donut/modules/arc.ts
```diff
@@ -45,7 +46,7 @@ export function updateArc<Datum> (
 ): void {
   selection
     .style('transition', `fill ${duration}ms`) // Animate color with CSS because we're using CSS-variables
-    .style('fill', (d, i) => getColor(d.data, config.color, i))
+    .style('fill', d => getColor(d.data, config.color, d.index))
```
rokotyan, line 49: > Shouldn't it be `d.data._index` or will they be the same?

@reb-dev Same as above

packages/ts/src/components/donut/types.ts
```diff
@@ -1,10 +1,14 @@
 import { PieArcDatum } from 'd3-shape'
 
+export type DonutInputArc<Datum> = Partial<DonutArcDatum<Datum>> & {
```
rokotyan, line 3: @reb-dev I'm a bit concerned about this type. From the code (see below) it looks like a wrapper around `Datum`
 and it doesn't have information about arcs (i.e. `startAngle`, `endAngle`, ...):
 
```ts
    const data: DonutInputArc<Datum>[] = datamodel.data.reduce((acc, d, i) =>
      !config.showEmptySegments && !getNumber(d, config.value, i)
        ? acc : [...acc, { _datum: d, _index: i }]
    , [])
```

So I would simply call it something like `DonutDatum` and define as:
```ts
export type DonutDatum<Datum> = Datum & {
  _index: number;
}
```
One can argue that original `Datum` can already have a property called `_index` and that will be a valid argument. However we already use such approach in many places in the library (e.g. in [Scatter](https://github.com/f5/unovis/blob/main/packages/ts/src/components/scatter/types.ts)).

packages/ts/src/components/donut/index.ts
```diff
@@ -57,38 +57,43 @@ export class Donut<Datum> extends ComponentCore<Datum[], DonutConfig<Datum>, Don
 
   _render (customDuration?: number): void {
     const { config, datamodel, bleed } = this
-    const data = datamodel.data
-    const duration = isNumber(customDuration) ? customDuration : config.duration
 
+    // Wrap data to preserve original indices
+    const data: DonutInputArc<Datum>[] = datamodel.data.reduce((acc, d, i) =>
```
rokotyan, line 62: @reb-dev I find this slightly more difficult to read comparing to the regular `map`  and `filter` approach.

```ts
const data: DonutDatum<Datum>[] = datamodel.data
  .map((d, i) => ({
    ...d,
    _index: i
  })
  .filter(d => !config.showEmptySegments && !getNumber(d, config.value, d._index))
```

P.S. See my comments regarding `DonutDatum` below

rokotyan (changes requested):

packages/ts/src/components/donut/types.ts
reb-dev: @rokotyan That was my original thought actually, but I don't think it will work for primitive types (i.e. number), so I think wrapping it is necessary. How do we bypass this in Scatter? Can we give data as an array of numbers?

packages/ts/src/components/donut/index.ts
reb-dev: @rokotyan I agree it's easier to read, but reduce is superior performance-wise so I usually prefer it. Your call though

packages/ts/src/components/donut/types.ts
rokotyan: @reb-dev Very good point!

It won't be possible to give data as an array of numbers to Scatter. Let's do the right thing for Donut and wrap it like you proposed in the beginning. Then there's no need to have the `_` prefix, right?
 
```ts
export type DonutDatum<Datum> = {
  datum: Datum;
  index: number;
}
```

packages/ts/src/components/donut/index.ts
rokotyan: @reb-dev Usually it is, but I don't think it's the case here:

```ts
[...acc, { _datum: d, _index: i }]
```
We're initializing a new array on every step, while memory allocation is an expensive operation which often becomes a performance bottleneck.

packages/ts/src/components/donut/modules/arc.ts
reb-dev: @rokotyan No actually, because the type `DonutArcDatum<Datum>` expects the `data` property to be the same type as our datum. In the previous code, after getting the arc datum we did this:

```
arcData.forEach((d, i) => {
  d.index = i
  d.innerRadius = innerRadius
  d.outerRadius = outerRadius
})
```

I tweaked it a bit to set d.data to `d.data_datum` and d.index to `d.data_index`. So `DonutDatum<Datum>` exists as a purely intermediate type. However would you prefer I preserve it? We could change the defintition to be `DonutDatum<Datum> extends PieArcDatum<DonuntDatum<Datum>>` ?

rokotyan: @reb-dev I see, thanks for explaining.

I did more thinking about it in the morning because the new type inheritance was making the code more difficult to read. Generally, I think that's the right approach. I pushed some changes that will make it slightly clearer in my opinion, i.e. I've added jsdoc comments to the types explaining what `index` means: 
```ts
  ...
  /** Original datum index as in unfiltered data */
  index: number;
```

Let me know if anything there doesn't look right to you. Also feel free to soft-reset the changes and do one single commit. Please don't forget to add a reference to the issue since it's missing in your original commit.

P.S. I've also added a config property for the empty segment size. I think it can be useful in some cases because with certain colors segments can be almost invisible at the default size of 0.5 degrees. Can you please update the docs accordingly?

reb-dev: @rokotyan everything looks good, thanks! I just updated the docs and did a single commit.

rokotyan (approved):

Issue #103, opened 2023-01-17
Tooltip triggers on timeline don't seem to work for line/lines and rows


mattp-:
forgive me, I don't have an easy to give a reproduction case via link/fiddle or whatever, but if you add 

+    3               [Timeline.selectors.line]: getTooltipText,                                                                                                                                                                                                                                                                                                              
+    4               [Timeline.selectors.lines]: getTooltipText       

to 
https://unovis.dev/gallery/view?collection=Lines%20and%20Areas&title=Basic%20Timeline
it does not seem to work as expected. Timeline.selectors.label seemingly does, but the line/lines selectors don't seem to trigger or install the mouseover handlers as I would expect.

mattp-:
poked at this a bit more, the problem seems to lie in 
https://github.com/f5/unovis/blob/main/packages/ts/src/components/tooltip/index.ts#L173
or atleast, that is why it is not working. on labels, the composedPath is 
```
0
: 
text.css-uter0n-label
1
: 
g.css-tqj2dj-labels
2
: 
g.css-1dybhmo-timeline-component
3
: 
svg.css-1q9y8bl
4
: 
div
...
```
the label is the first element in the path. on rows/lines, the composedPath() of the event is
```
0
: 
rect.css-1b6t1i9-background
1
: 
g.css-1dybhmo-timeline-component
2
: 
svg.css-1q9y8bl
3
: 
div
...
```
the background g set up by timeline. I'm not too familiar with how svg works with event handlers / why this behavior is different for the labels and not the rows/lines. hopefully this should help isolate the problem at least.

rokotyan:
@mattp- Thanks for posting this, it's clearly a bug! Adding to our to do list.

rokotyan:
@mattp- The fix will be released in version `1.0.3` but you can already use version `1.0.3-beta.0` if you want

mattp-:
@rokotyan im seeing a similar issue even with the referenced PR changes when trying to use  VisTimeline events={[Timeline.selectors.rows]: ... } , it seems like the same / similar issue. should I open another issue?

rokotyan:
@mattp- Can you please post a screenshot of the error here?

mattp-:
no errors, just when i try to click a row (not the bar itself which works), it does not fire the event handler; ive got both
Timeline.selectors.rect and Timeline.selectors.rows in the events dict.

rokotyan:
@mattp- I was able to reproduce the issue, it will be fixed it in the next update

[...]

PR #138, opened 2023-02-14
Component | Tooltip: Changing the way we handle scrolling

rokotyan:
Fixes [#103](https://github.com/f5/unovis/issues/103)

• Enabling `pointer-events` for rows to allow for tooltip events;
• Attaching row type (label) as data to row elements;
• Removing the logic of setting `pointer-events` to `false` while scrolling and adding `this._onMouseWheel` event handler to `[Timeline.selectors.line]`. The tooltip will be hidden upon scrolling.

https://user-images.githubusercontent.com/755708/218609339-4e824eb4-409b-4a2a-837e-f47bf2907907.mp4


reb-dev (approved):
Everything looks good!

Issue #141, opened 2023-02-16
Component | Graph: ELK Layout support


rokotyan:
https://www.eclipse.org/elk/reference/algorithms/org-eclipse-elk-layered.html

PR #161, opened 2023-03-13
Component | Graph: ELK Support

rokotyan:
This PR implements initial ELK support, which is the best layout engine for diagrams as of today. ELK Layered Layout can be a good alternative for our Parallel Layout when you need more nesting (panels within panels ...). Besides that ELK has [other  layouts](https://www.eclipse.org/elk/reference/algorithms.html) available. #141

SCR-20230313-otnw
SCR-20230313-otfu

image


packages/website/docs/networks-and-flows/Graph.mdx
```diff
@@ -432,6 +432,79 @@ object to `dagreLayoutSettings`.
...
+  switch (layer) {
+    case 'root': return { ... }
+    case 'group1': return { ... }
+    ...
+  }
+}
+```
+
+The ELK Layered algorithm works best when combined with [panels](#panels), allowing you to highlight the hierarchical
+nature of your graph data. Check out the ELK example in the gallery to learn more.
```
reb-dev, line 461: > Check out the ELK example in the gallery to learn more.

Maybe add a hyperlink?

packages/ts/src/components/graph/modules/panel/helper.ts
```diff
@@ -26,26 +26,37 @@ export const DEFAULT_LABEL_MARGIN = 16
...
+  return {
+    left: (isPaddingAnObject ? (padding as Spacing).left : (padding as number)) ?? DEFAULT_PADDING,
+    right: (isPaddingAnObject ? (padding as Spacing).right : (padding as number)) ?? DEFAULT_PADDING,
+    top: (isPaddingAnObject ? (padding as Spacing).top : (padding as number)) ?? DEFAULT_PADDING,
+    bottom: (isPaddingAnObject ? (padding as Spacing).bottom : (padding as number)) ?? DEFAULT_PADDING,
+  }
+}
+
+export function initPanels (panelsConfig: GraphPanelConfig[]): GraphPanel[] {
+  const panels = panelsConfig.map(p => ({
```
reb-dev, line 40: We need to change it to `panelsConfig?.map` or set a default argument here because right now all of the graphs are broken when `panels` is not provided to the config.

Screen Shot 2023-03-17 at 12 21 50 PM

reb-dev (commented):
This is really cool! Nice work.

One thing I'm not sure about: is there a straight forward way to change the layout algorithm? Because when I change the dev example's props to `layoutElkSettings={{ 'elk.algorithm': 'radial' }}` I get an error, but inside the component itself I can just add that to the layout options and it works:

Screen Shot 2023-03-17 at 12 31 56 PM
Screen Shot 2023-03-17 at 12 31 34 PM

packages/website/docs/networks-and-flows/Graph.mdx
rokotyan: Good point. Added.

packages/ts/src/components/graph/modules/panel/helper.ts
rokotyan: I can't believe I missed that, thanks for pointing out! Fixed

rokotyan:
> One thing I'm not sure about: is there a straight forward way to change the layout algorithm? Because when I change the dev example's props to `layoutElkSettings={{ 'elk.algorithm': 'radial' }}` I get an error, but inside the component itself I can just add that to the layout options and it works

@reb-dev Thanks! ELK has a quite sophisticated layout configuration logic that I've tried to wrap it a little to suit our primary use case out of the box. But it looks like it broke some other things, so I've simplified my logic to avoid issues like the one you've found.

reb-dev (approved):
Looks good!

PR #162, opened 2023-03-14
React | Tweaks: Supporting `style` and `className`; Removing `@emotion/css` dependency

rokotyan:
A couple of minor tweaks:
* Supporting `style` and `className` in containers and html components
* Removing `@emotion/css` dependency to avoid installation warnings of unmet peer-dependencies


reb-dev (approved):
Looks good.

PR #170, opened 2023-03-30
New Component: Nested Donut

reb-dev:
#172


rokotyan (commented):
@reb-dev Great start! Everything looks good so far, I just left a couple comments.

A few general thoughts:
1. I think the component should support direction (inwards or outwards) and be able to look like a sunburst chart when the hierarchy is uneven (i.e. when different sections have different maximum depth): 
image
2. By default, let's apply the shades of the same color for segments of the same section, like here:  image
3. I would also like to add labels to segments, but let's do it on a later stage.

packages/ts/src/components/nested-donut/index.ts
```diff
@@ -0,0 +1,162 @@
...
+    this.centralLabel = this.g.append('text')
+      .attr('class', s.centralLabel)
+    this.centralSubLabel = this.g.append('text')
+      .attr('class', s.centralSubLabel)
+  }
+
+  private getHierarchyData (): NestedDonutSegment<Datum>[] {
+    const { config, datamodel } = this
+
+    const layerAccessors = config.layers.map(layerAccessor => (d: Datum) => getString(d, layerAccessor))
```
rokotyan, line 62: @reb-dev If there's a way to provide the datum index here, I think we should do it: `getString(d, layerAccessor, i)`

rokotyan (commented):
@reb-dev Overall looks good! I've left a few comments to the code, and there's a problem with color transitions (there's some flickering):
https://user-images.githubusercontent.com/755708/230653417-660172cc-d739-4d0f-bb1b-4577d2d2ebb2.mp4

Since the majority of data vis issues come in edge cases, can you please think how we can incorporate them into the dev app? In our old demo we used to have separate visualization states on one page:  "regular", "no data", "single data element", "a lot of data", and transitions between them. Maybe we can do something similar here.

rokotyan:
@reb-dev The component looks solid, great job! A few UX/UI comments first:
* I would add `user-select: none` to the labels to prevent things like this:
SCR-20230502-lmnf

* It looks like the label coloring logic (dark/bright) doesn't work on the example page above;
* I think we also need to implement some additional logic to check whether the labels fit into their segments or not. Those that don't fit will be hidden by default.

 I would also like us to work more on transitions:
* Currently it looks like the update transition has some artifacts:

https://user-images.githubusercontent.com/755708/235774845-12919b1a-ff8d-4129-8386-9b4066b0cdc6.mp4

* The enter / exit transition can be slightly more dynamic (currently it only uses opacity):

https://user-images.githubusercontent.com/755708/235774861-43abafe2-620a-4e6d-a78b-a056b185ae68.mp4

Do you feel like you know how to tweak them or would you like my help? We can hop on a call and discuss different options.

packages/ts/src/components/nested-donut/index.ts
```diff
@@ -0,0 +1,197 @@
...
+    if (config) this.config.init(config)
+    this.arcBackground = this.g.append('g')
+    this.arcGroup = this.g.append('g')
+    this.centralLabel = this.g.append('text')
+      .attr('class', s.centralLabel)
+    this.centralSubLabel = this.g.append('text')
+      .attr('class', s.centralSubLabel)
+  }
+
+  private getHierarchyData (layers: NestedDonutLayer[]): NestedDonutSegment<Datum>[] {
```
rokotyan, line 64: I don't have a super strong option on it, but I prefer to prefix private methods with an underscore `_`. This makes it easier to read through the code because you can immediately see which method calls are purely internal.

packages/ts/src/components/nested-donut/style.ts
```diff
@@ -0,0 +1,82 @@
...
+
+const cssVarDefaults = {
+  '--vis-nested-donut-background-color': '#E7E9F3',
+  // Undefined by default to allow proper fallback to var(--vis-font-family)
+  '--vis-nested-donut-font-family': undefined,
+
+  // Central label
+  '--vis-nested-donut-central-label-font-size': '16px',
+  '--vis-nested-donut-central-label-font-weight': 600,
+  '--vis-nested-donut-central-label-text-color': labelColors.dark,
```
rokotyan, line 19: I would also add CSS variables for dark / light segment labels, like we have it in Graph or LeafletMap:
```
'--vis-map-point-inner-label-text-color-dark': '#5b5f6d',
'--vis-map-point-inner-label-text-color-light': '#fff',
```

it looks like now there's no way to change default `labelColors` for segments.

rokotyan (changes requested):

reb-dev:
@rokotyan I just updated the branch with those style tweaks as well as the label rendering logic. I think the sizing logic could use some more tweaking, but I think it looks decent enough for now.

Will update you when I've made progress on transitions. Might reach out on Slack tomorrow if I get stuck.

rokotyan:
@reb-dev Sounds good! Btw, I think we should make labels hiding configurable.

rokotyan (approved):

PR #171, opened 2023-03-30
Component | LeafletMap: POC of having an object of CSS variable names

rokotyan:
For a long time I was concerned about the names of CSS variables in Unovis: we reference them in the TS code and every time the name of a certain variable changes, we need to manually change it everywhere else

This PR is a proof of concept of having the names of component's CSS variables stored in an object, that can be used in the code and can be accessed externally as a static class property.


reb-dev (approved):
Nice! I like this change.

PR #174, opened 2023-03-31
Container | Core, Single: Setting SVG size in `render()`

rokotyan:
Setting SVG size in `render()` to avoid unnecessary rendering caused by ResizedObserve, that also interrupted active transitions causing some of the components look "stuck".

### Before
image
image

### After
image
image

#173

### DIFF
--- a/packages/ts/src/components/sankey/index.ts
+++ b/packages/ts/src/components/sankey/index.ts
@@ -219,7 +219,7 @@ export class Sankey<
 
     const height = max(values) || config.nodeMinHeight
     this._extendedHeight = height + bleed.top + bleed.bottom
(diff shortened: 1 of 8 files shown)


reb-dev (changes requested):
@rokotyan Looks good except I'm getting console errors when opening the sankey dev example:
Screen Shot 2023-04-03 at 11 34 25 AM

rokotyan:
> @rokotyan Looks good except I'm getting console errors when opening the sankey dev example

@reb-dev Thanks for reporting! It allowed me to find a few more rendering issues. Fixed in the latest commits.

reb-dev (approved):

PR #179, opened 2023-04-04
Component | Stacked Bar, Groped Bar: Smarter bleed calculations

### LINKED ISSUE #178: Stacked Bar, Grouped Bar: Bleed spacing is being applied when it's unnesessary
When `xDomain` is set to a range that fits all the bars, the bleed spacing is still being applied:
image

rokotyan:
Fixes #178 
> When xDomain is set to a range that fits all the bars, the bleed spacing is still being applied:

### Before
image

### After
image


reb-dev (changes requested):
@rokotyan I noticed an edge case. When adding a data point that should be within the domain, (in this case `{ x: -1, y: 10, y1: 5, y2: 1 }`), the whole group isn't shown:

Screen Shot 2023-04-07 at 11 11 48 AM

rokotyan:
@reb-dev Great catch! The coded was not able to handle unsorted data. Fixed in the latest commit.

reb-dev (approved):

Issue #180, opened 2023-04-05
Website | Gallery: Expandable Sankey example not working properly


rokotyan:
https://user-images.githubusercontent.com/755708/229948560-bf5d5403-ae47-479b-ab57-7a0ff69b7286.mp4

PR #185, opened 2023-04-22
Website | Gallery: Fix expandable sankey data model

reb-dev:
#180

**Background:**
The error was occurring because `data.nodes` and `data.links` weren't in sync. Previously, only the `node` objects were updated when setting a node's `expanded` property, while `links` remained unchanged, resulting in links with unrecognized source and target values when _Sankey_ wasn't fully expanded. (I suspect changes from [Datamodel | Graph: Fix link data logic (8b4416d) ](https://github.com/f5/unovis/commit/8b4416d218106340c6b3dcb998fcf7cc6ffeb8e6#diff-b4e9556c6675215c7be1e5f1db3804a7450e0e341c40e2d4dfcd11c908fa5333) are what ultimately broke it)

 **Changes** :
This PR:
- Updates both nodes and links whenever a node's subtree is toggled
- Extracts expand/collapse logic to `data.ts` file to maintain clarity in the component code
- Changes the data shape and type defs a bit to be more readable
- Includes minor tweaks to increase consistency across example files

### DIFF
--- a/packages/website/src/examples/expandable-sankey/data.ts
+++ b/packages/website/src/examples/expandable-sankey/data.ts
@@ -1,260 +1,132 @@
-export type NodeDatum = {
+export type Node = {
   id: string;
   label: string;
-  subgroups: string[];
   value: number;
+  subgroups: Node[];
+  color?: string;
   expandable?: boolean;
   expanded?: boolean;
-  level?: number;
-};
+}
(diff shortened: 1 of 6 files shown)


rokotyan (approved):
@reb-dev Thanks for fixing the problem and explaining the reason, appreciate it.

PR #189, opened 2023-04-26
Dev Demo | Adding data transition pages

reb-dev:
#0

For every component in the `examples` directory, we can export default type of `TransitionComponentProps` to render the data transition page. This type includes the base component (which accepts a single data prop), the default data generator function, and a collection of data to test. On the page itself, we can toggle the animation by clicking play/pause or reset the generated data by clicking refresh.

I added some basic pages for Donut, Sankey, Stacked Bar, Area and Timeline to get started.

https://user-images.githubusercontent.com/52078477/234433618-6a0caed6-4249-47e1-8e2f-b5ce075baded.mov


packages/dev/src/examples/index.ts
```diff
@@ -17,12 +19,19 @@ export type ExampleItem = {
...
-    category: module.category,
-    component: module.component,
+  const category = key.match(/\.\/[\w-]+\/([\w-]+)\//)?.[1] as string
+
+  const exampleItem = {
+    ...module,
+    category: kebabToTitleCase(category),
+  }
+
+  if (module.default) {
```
rokotyan, line 29: While this approach works, I'm not sure whether it's a good pattern. What if we exported transition components as named exports with a certain name? For example `module.transitionComponent`. It will allow us to introduce other component types in the future if we need them.

rokotyan (approved):
@reb-dev Great job! I really like how it looks and how it's engineered. The only concern that I have is related to using the `default` export to identify Transition Components. See my comment below.

packages/dev/src/examples/index.ts
reb-dev: Good point! I've changed default to `transitionComponent`.

rokotyan (approved):

Issue #194, opened 2023-05-05
Core | Container: Resize Observer doesn't trigger a resize if it happens right after the first animation frame


rokotyan:
If the parent element containing the chart has zero width and height at the initialization phase, and then it is resized, the chart won’t be updated.

PR #195, opened 2023-05-05
Container: Fixing Reize Observer configuration

rokotyan:
Fix for #194

Set up a Resize Observer in ﻿render() instead of ﻿_render() to capture any changes to the container size that might occur in the next animation frame after the initial ﻿render() is called.


reb-dev (approved):
@rokotyan Approving for logic. However I couldn't reproduce #194 on main, how did you discover the issue?

rokotyan:
@reb-dev It popped up in F5XC, I think it's an Angular-specific thing.

PR #205, opened 2023-05-19
Component | LeafletMap: Add getExpandedCluster public method

reb-dev:
#0

### DIFF
--- a/packages/ts/src/components/leaflet-map/index.ts
+++ b/packages/ts/src/components/leaflet-map/index.ts
@@ -386,6 +386,14 @@ export class LeafletMap<Datum extends GenericDataRecord> extends ComponentCore<D
     this.render()
   }
 
+  /** Get the currently expanded cluster */
+  public getExpandedCluster (): { id: string; points: Datum[] } | undefined {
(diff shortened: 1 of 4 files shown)


packages/ts/src/components/leaflet-map/index.ts
```diff
@@ -386,6 +386,14 @@ export class LeafletMap<Datum extends GenericDataRecord> extends ComponentCore<D
     this.render()
   }
 
+  /** Get the currently expanded cluster */
+  public getExpandedCluster (): { id: string; points: Datum[] } {
+    if (!this._expandedCluster) return
```
rokotyan, line 391: @reb-dev Since the function can also return `undefined` let's update the return type to `{ id: string; points: Datum[] } | undefined`

rokotyan (changes requested):

packages/ts/src/components/leaflet-map/index.ts
reb-dev: Fixed!

rokotyan (approved):

PR #211, opened 2023-05-25
Nested Donut: Fixes and doc revisions

reb-dev:
This PR:
- Fixes issue where selector data wouldn't show correct values after update (needed to change data ids)
- Adds exit transition for background layers
- Updates the _Segment Values_ dev example to showcase the two above fixes
- Various doc tweaks

### DIFF
--- a/packages/website/docs/misc/NestedDonut.mdx
+++ b/packages/website/docs/misc/NestedDonut.mdx
@@ -1,6 +1,9 @@
+---
+title: Nested Donut
+---
 import { PropTable } from '../components/props-table.tsx'
 import { sample } from '../utils/data.ts'
-import { DocWrapper, InputWrapper } from '../wrappers'
+import { DocWrapper } from '../wrappers'
 
 export const data = Array(100)
   .fill(0)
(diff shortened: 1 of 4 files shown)


packages/website/docs/misc/NestedDonut.mdx
```diff
@@ -87,67 +68,98 @@ layers = [(d: Datum) => d.group, (d: Datum) => d.value]
...
+```ts
+const data = [{ label: 'A', value: 3 }, { label: 'B', value: 2 }, { label: 'C', value: 1 }]
+```
+
+you can provide `value` with `(d: Datum) => d.value` to acheive the same effect.
 
-There are custom settings you can set for each individual layer. Just provide `layerSettings` property
-with an accessor that returns the following type based on the layer's depth:
+## Layer Settings
+You can customize the appearance for each inidividual layer with the `layerSettings` property,
```
rokotyan, line 99: inidividual → individual

packages/website/docs/misc/NestedDonut.mdx
```diff
@@ -87,67 +68,98 @@ layers = [(d: Datum) => d.group, (d: Datum) => d.value]
...
+  ...Array(1).fill('C'),
+  ]}
+  layerSettings={{ labelAlignment: 'along' }}
+  layers={[d => d]}
+  excludeTabs
 />
 
-### Layer Settings
+### Custom `value` Accessor
+Alternatively, you can override the default way _Nested Donut_ calculates values by providing your own `value` accessor funciton.
```
rokotyan, line 90: funciton → function

rokotyan (approved):

Issue #202, opened 2023-05-16
document is not defined (SvelteKit)


ItsBioDude:
So here is the issue
```powershell
ReferenceError: document is not defined
    at new ComponentCore (file:///D:/Hobby/thuchanh/node_modules/.pnpm/@unovis+ts@1.1.1/node_modules/@unovis/ts/core/component/index.js:25:28)
    at new XYComponentCore (file:///D:/Hobby/thuchanh/node_modules/.pnpm/@unovis+ts@1.1.1/node_modules/@unovis/ts/core/xy-component/index.js:9:9)
    at new Line (file:///D:/Hobby/thuchanh/node_modules/.pnpm/@unovis+ts@1.1.1/node_modules/@unovis/ts/components/line/index.js:16:9)
    at eval (/node_modules/.pnpm/@unovis+svelte@1.1.1_@unovis+ts@1.1.1_svelte@3.59.1/node_modules/@unovis/svelte/components/line/line.svelte:26:20)
    at Object.$$render (/node_modules/.pnpm/svelte@3.59.1/node_modules/svelte/internal/index.mjs:2042:22)  
    at Object.default (/src/routes/tara/+page.svelte:135:100)
    at eval (/node_modules/.pnpm/@unovis+svelte@1.1.1_@unovis+ts@1.1.1_svelte@3.59.1/node_modules/@unovis/svelte/containers/xy-container/xy-container.svelte:72:156)
    at Object.$$render (/node_modules/.pnpm/svelte@3.59.1/node_modules/svelte/internal/index.mjs:2042:22)  
    at Object.default (/src/routes/tara/+page.svelte:133:103)
    at eval (/src/lib/StatisticPage.svelte:22:33)
    at Object.$$render (/node_modules/.pnpm/svelte@3.59.1/node_modules/svelte/internal/index.mjs:2042:22)  
    at eval (/src/routes/tara/+page.svelte:114:170)
    at Object.$$render (/node_modules/.pnpm/svelte@3.59.1/node_modules/svelte/internal/index.mjs:2042:22)  
    at Object.default (/.svelte-kit/generated/root.svelte:74:138)
    at eval (/node_modules/.pnpm/@sveltejs+kit@1.16.3_svelte@3.59.1_vite@4.3.5/node_modules/@sveltejs/kit/src/runtime/components/layout.svelte:8:41)
    at Object.$$render (/node_modules/.pnpm/svelte@3.59.1/node_modules/svelte/internal/index.mjs:2042:22)  
[...]

Issue #177, opened 2023-04-03
Issue with css dependency: "process not defined". 


echobops:
I'm following the basic Svelte install / demo instructions and have come across an error with one of the depencies (caught in the browser): 

```
emotion-serialize.browser.esm.js:47 Uncaught ReferenceError: process is not defined
    at emotion-serialize.browser.esm.js:47:1
    at main.js:7:2
```

This is from the @emotion/css package. I'm not sure why its trying to access the "process" variable in the browser. I realize this isn't your package, but am not sure where else to post the issue. I'm using:

rollup: ^2.79.1
svelte: ^3.58.0

rokotyan:
@echobops Can you please share more details about your Svelte set up (e.g. are you using SvelteKit)?

echobops:
I am using "uibuilder" (node-red-contrib-uibuilder v6.3.0), an NPM module for Node-red. It built a project for me, which has the following package.json (with some additions there by myself). Normally I'd add @unovis/svelte /ts as dependancies here. 

```
{
  "name": "svelte-basic",
  "description": "A basic Svelte template for uibuilder and Node-RED.",
  "version": "1.0.0",
  "private": true,
  "scripts": {
    "build": "rollup -c --environment NODE_ENV:production",
    "dev": "rollup -c -w --environment NODE_ENV:development",
    "start": "sirv public --no-clear"
  },
  "devDependencies": {
    "@rollup/plugin-commonjs": "^17.0.0",
    "@rollup/plugin-node-resolve": "^11.0.0",
    "locally-unique-id-generator": "^0.1.5",
    "rollup": "^2.79.1",
    "rollup-plugin-css-only": "^3.1.0",
    "rollup-plugin-livereload": "^2.0.0",
    "rollup-plugin-svelte": "^7.0.0",
    "rollup-plugin-terser": "^7.0.0",
    "svelte": "^3.58.0"
  },
  "dependencies": {
    "feather-icons": "^4.29.0",
    "lodash": "^4.17.21",
    "luxon": "^3.2.1",
    "navigo": "^8.11.1",
    "sirv-cli": "^2.0.0",
    "svelte-frappe-charts": "^1.10.0",
    "svelte-octicons": "^17.13.0",
    "sveltestrap": "^5.9.0"
  }
}
```

rokotyan:
Thanks @echobops, as a quick test can you please try initializing Unovis's components after the component has been mounted? Here's an example:

```svelte
<script>
[...]

PR #216, opened 2023-05-30
@unovis/svelte: Update component lifecycles to prevent DOM related errors

reb-dev:
#202, #177


rokotyan:
@reb-dev I've been testing you PR with SvelteKit and found that there's one more error that was introduced with my recent ResizeObserver changes:
https://github.com/f5/unovis/blob/6dcb9130dcab69f1c9d3ddbe684a76ccd7cbf350/packages/ts/src/utils/resize-observer.ts#L2

To fix it we'll just need to replace `window` with `globalThis`. Can you please add this change to this PR?

reb-dev:
@rokotyan 
> To fix it we'll just need to replace window with globalThis. Can you please add this change to this PR?

Done!

rokotyan (approved):

Issue #228, opened 2023-06-08
Component | Chord Diagram: Initial render doesn't use correct `bleed` value


reb-dev:
After resizing, we see it adjusts to the expected size:

https://github.com/f5/unovis/assets/52078477/3508aa68-42d2-4c35-92d2-6d7831efab64

PR #235, opened 2023-06-15
Component | Chord Diagram: Fix bleed logic + minor refactoring

reb-dev:
#228


rokotyan:
@reb-dev Had a quick look and spotted this weird behavior:


https://github.com/f5/unovis/assets/755708/6ff80c2c-1ecb-461a-9138-d216aab511f7

reb-dev:
@rokotyan You're right. I realized I had to rewrite the logic for link rendering so that the current bleed value is applied (before it was precalcualted in `_getRibbons`). It should look normal now 👍

rokotyan:
@reb-dev Thanks! Let's include this into Unovis 1.2 too. 

I would like us to do some more tweaks to the component, but let's do it later when we finalize the new design requirements.

rokotyan (approved):

PR #263, opened 2023-08-18
Add Auto Generated Aria Description For XY Container

ivegotwings:
Support for Autogenerated Aria-Label for xy-container

Screenshot 2023-08-18 at 12 56 10 PM


packages/ts/src/core/container/index.ts
```diff
@@ -61,9 +59,13 @@ export class ContainerCore {
...
 
-    // Apply the `aria-label` attribute
-    select(this._container)
-      .attr('aria-label', config.ariaLabel)
+    // use setTimout to allow _render to complete in the derived class
+    setTimeout(() => {
+      // Apply the `aria-label` attribute
+      select(this._container)
+        .attr('aria-label', this.getAriaDescription())
+        .attr('tabindex', '0')
```
ivegotwings, line 67: note this

rokotyan: @ivegotwings Can you please explain in more detail whey do we need to delay rendering here?

packages/ts/src/components/grouped-bar/index.ts
```diff
@@ -1,29 +1,25 @@
-import { scaleBand } from 'd3-scale'
```
rokotyan, line 1: The import order got messed up in all of the edited files

packages/ts/src/components/grouped-bar/index.ts
```diff
@@ -322,4 +318,9 @@ export class GroupedBar<Datum> extends XYComponentCore<Datum, GroupedBarConfig<D
       ? this.getDataScaleExtent()
       : this.getValueScaleExtent(false)
   }
+
+  getAriaDescription (): string {
+    const description = super.getAriaDescription() || ''
```
rokotyan, line 323: I'm not sure that calling `super.getAriaDescription()` is a good idea. If you have multiple components in your chart, you'll most  likely end up with having duplicated items in the description.

packages/ts/src/containers/xy-container/index.ts
```diff
@@ -50,6 +46,7 @@ export class XYContainer<Datum> extends ContainerCore {
   private _clipPathId = guid()
   private _axisMargin: Spacing = { top: 0, bottom: 0, left: 0, right: 0 }
   private _firstRender = true
+  private xyDomain = new Map<ScaleDimension, number[]>()
```
rokotyan, line 49: We prefix all private variables with underscore `_`. The rule is not enforced with ESLint yet, but we follow it.

packages/ts/src/core/component/index.ts
```diff
@@ -151,4 +147,8 @@ export class ComponentCore<
   public isDestroyed (): boolean {
     return !this.element
   }
+
+  public getAriaDescription (): string {
```
rokotyan, line 151: I think it's worth adding a comment about this method and why it return an empty string here in the base class

packages/ts/src/core/container/config.ts
```diff
@@ -34,6 +33,8 @@ export interface ContainerConfigInterface {
    * `aria-label` attribute to the div element containing your chart. Default: `undefined`.
   */
   ariaLabel?: string | null | undefined;
+
+  ariaAutoLabel?: boolean;
```
rokotyan, line 37: Need a description

packages/ts/src/core/container/config.ts
```diff
@@ -58,4 +59,5 @@ export class ContainerConfig extends Config implements ContainerConfigInterface
 
   svgDefs = undefined
   ariaLabel = undefined
+  ariaAutoLabel = false
```
rokotyan, line 62: Let's set it to `true` by default

packages/ts/src/core/container/index.ts
```diff
@@ -148,4 +150,10 @@ export class ContainerCore {
     this._resizeObserver?.disconnect()
     this.svg.remove()
   }
+
+  getAriaDescription (): string {
```
rokotyan, line 154: `getAriaDescription` → `protected _getAriaDescription`

packages/ts/src/core/container/index.ts
rokotyan: Also let's handle the logic related `ariaAutoLabel` and `ariaLabel` configuration here, not in the `getAriaDescription` method. See my comment below.

packages/ts/src/core/container/index.ts
rokotyan: I find it a bit difficult to understand how this method will work and I don't think it's going to work as supposed:

When you call `select(this._container).attr('aria-label', this.getAriaDescription())`, it'll first call the `getAriaDescription` method of the container instance, which in turn will call `super.getAriaDescription` that will return `config.ariaLabel` if provided. So you'll end up having the provided and generated descriptions at the same time.

Let's separate the description generation logic from the rendering logic. It will make it easier to understand the code.

packages/ts/src/core/xy-component/index.ts
```diff
@@ -87,4 +84,12 @@ export class XYComponentCore<
     const yAccessors = (isArray(config.y) ? config.y : [config.y]) as NumericAccessor<Datum>[]
     return getExtent(data, ...yAccessors)
   }
+
+  getAriaDescription (): string {
+    const xDataExtent = this.getXDataExtent()
+    const yDataExtent = this.getYDataExtent(true)
+    let description = `Extent of ${ScaleDimension.X} is ${xDataExtent[0]?.toFixed(2)} to ${xDataExtent[1]?.toFixed(2)} `
```
rokotyan, line 91: `Extent of ${ScaleDimension.X} is ${xDataExtent[0]?.toFixed(2)} to ${xDataExtent[1]?.toFixed(2)} ` → 
`Extent of the ${ScaleDimension.X} dimension spans from ${xDataExtent[0]?.toFixed(2)} to ${xDataExtent[1]?.toFixed(2)}`

If the extent is `undefined`, the description will say "undefined to undefined" . I think that edge case has to be handled separately, and say something like: "The extent of the X dimension is undefined"

packages/ts/src/containers/xy-container/index.ts
```diff
@@ -408,4 +406,22 @@ export class XYContainer<Datum> extends ContainerCore {
...
   }
+
+  getAriaDescription (): string {
+    const ariaLabel = super.getAriaDescription()
+    if (this.config.ariaAutoLabel) {
+      let description = ''
+      // Get X and Y Domain
+      Object.values(ScaleDimension).forEach((dimension: ScaleDimension) => {
+        const domain = this.xyDomain[dimension]
+        description += `${dimension} domain ${domain?.[0]?.toFixed(2)} to ${domain?.[1]?.toFixed(2)} `
```
rokotyan, line 417: `${dimension} domain ${domain?.[0]?.toFixed(2)} to ${domain?.[1]?.toFixed(2)} ` →
`${dimension} scale domain is from ${domain?.[0]?.toFixed(2)} to ${domain?.[1]?.toFixed(2)} `

packages/ts/src/containers/xy-container/index.ts
```diff
@@ -408,4 +406,22 @@ export class XYContainer<Datum> extends ContainerCore {
     xAxis?.destroy()
     yAxis?.destroy()
   }
+
+  getAriaDescription (): string {
+    const ariaLabel = super.getAriaDescription()
+    if (this.config.ariaAutoLabel) {
```
rokotyan, line 412: Pls See my comment about logic separation below

packages/ts/src/core/container/index.ts
ivegotwings: We need this to complete the execution of `_render` function in the xy-container so as to generate the value of `this.xyDomain`which is later used in `getAriaDescription` 

If we do not use the delay `getAriaDescription` gets called before `the _render` method completes and the `this.xyDomain` remains undefined

packages/ts/src/components/grouped-bar/index.ts
ivegotwings: I was under the impression that each XY-component will generate a unique description as they will have a unique x extent and y extent ?

packages/ts/src/core/component/index.ts
ivegotwings: The only requirement for this method to be present in the base class is basically to support inheritance, so that we can call the derived class method via polymorphism

packages/ts/src/core/container/index.ts
ivegotwings: As per the updated logic, we generate description like so-
`const description = config.ariaAutoLabel ? this._getAriaDescription() : config.ariaLabel`

packages/ts/src/components/grouped-bar/index.ts
```diff
@@ -322,4 +322,9 @@ export class GroupedBar<Datum> extends XYComponentCore<Datum, GroupedBarConfig<D
       ? this.getDataScaleExtent()
       : this.getValueScaleExtent(false)
   }
+
+  getAriaDescription (): string {
```
rokotyan, line 326: To add `public` keyword

packages/ts/src/core/container/config.ts
```diff
@@ -34,6 +34,11 @@ export interface ContainerConfigInterface {
    * `aria-label` attribute to the div element containing your chart. Default: `undefined`.
   */
   ariaLabel?: string | null | undefined;
+
+  /**
+   * Set to true to generate automated description for aria-label in containers
```
rokotyan, line 39: Add the `default` value.

packages/ts/src/core/xy-component/index.ts
```diff
@@ -87,4 +87,23 @@ export class XYComponentCore<
...
   }
+
+  getAriaDescription (): string {
+    const xDataExtent = this.getXDataExtent()
+    const yDataExtent = this.getYDataExtent(true)
+    let description
+    if (xDataExtent[0] === undefined && xDataExtent[1] === undefined) {
+      description = 'The extent of the X dimension is undefined '
+    } else {
+      description = `Extent of the ${ScaleDimension.X} dimension spans from ${xDataExtent[0]?.toFixed(2)} to ${xDataExtent[1]?.toFixed(2)} `
```
rokotyan, line 98: To apply text formatting (dots, trimming the text, capitalizing X and Y)

packages/ts/src/core/container/index.ts
```diff
@@ -61,9 +61,15 @@ export class ContainerCore {
       this.svg.append('defs').attr('class', 'svgDefs').html(config.svgDefs)
     }
 
-    // Apply the `aria-label` attribute
-    select(this._container)
-      .attr('aria-label', config.ariaLabel)
+
+    // use setTimout to allow _render to complete in the derived class
+    setTimeout(() => {
+      const description = config.ariaAutoLabel ? this._getAriaDescription() : config.ariaLabel
```
rokotyan, line 67: `config.ariaLabel` should have precedence

rokotyan (commented):
Let's also add `ariaAutoLabel` and `ariaLabel` description configuration to the component.

rokotyan:
Also let's add Axis labels description

rokotyan:
@ivegotwings I've prepared a PR that introduces a new preRender step and pushed a few changes to your branch. Please take a look and let me know what you think.

ivegotwings:
> @ivegotwings I've prepared a PR that introduces a new preRender step and pushed a few changes to your branch. Please take a look and let me know what you think.

Thanks @rokotyan

rokotyan:
Thanks for the updates @ivegotwings, looks good! Now, I think we need to work on the actual description copy to make it easier to perceive. I'll follow up on that

reb-dev (changes requested):
Good progress overall. In addition to my other comments, the main thing I would change is the language in the descriptions themselves. Let's refer this table (only appending the last column when `k > 1`):

|     | Base description<br/>(_n_ =`data.length`) | With multi-dimensional data<br/>(_k_ = `yAccessors.length`) |
| ---  | :----- | :-------------------- | 
| `Area` | Area chart | with _k_ stacked areas |
| `Line` | Line chart |  with _k_ lines |
| `Scatter` | Scatter plot  with _n_ point(s) |
| `GroupedBar` | Bar chart with _n_ item(s) | , each containing _k_ bars grouped together |
| `StackedBar` | Bar chart with _n_ item(s) |  , each containing _k_ bars stacked vertically* |
| `Timeline` | Timeline with _n_ segment(s) | placed in _k_ rows** |

\* - or "horizontally" depending on `config.orientation`</span>
\*\* - if `config.showLabels !== false`, we should mention the labels too. i.e. _Timeline with 20 segments placed in 4 rows labeled "A", "B", "C", and "D"._

packages/ts/src/containers/xy-container/index.ts
```diff
@@ -408,4 +416,33 @@ export class XYContainer<Datum> extends ContainerCore {
...
+
+      const domain = this._scaleDomains[dimension]
+      const axisLabel = this.config?.[`${dimension}Axis`]?.config.label
+      description += axisLabel
+        ? `The ${
+          dimension === ScaleDimension.X ? 'Horizontal Axis' : 'Vertical Axis'
+        } is labeled as ${axisLabel} and `
+        : `The ${
+          dimension === ScaleDimension.X ? 'Horizontal Axis' : 'Vertical Axis'} `
+      description += `ranges from ${dimension === ScaleDimension.X ? tickFormat(+domain?.[0]?.toFixed(2)) : domain?.[0]?.toFixed(2)} to ${dimension === ScaleDimension.X ? tickFormat(+domain?.[1]?.toFixed(2)) : domain?.[1]?.toFixed(2)}. `
```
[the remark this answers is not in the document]
rokotyan: @reb-dev I agree with your comments. I've a few more concerts about the overall architecture of the feature, will post them below.

packages/ts/src/core/container/index.ts
```diff
@@ -62,15 +68,24 @@ export class ContainerCore {
     }
 
     // Apply the `aria-label` attribute
-    select(this._container)
-      .attr('aria-label', config.ariaLabel)
+    let description = ''
+    if (config.ariaLabel) {
+      description = config.ariaLabel
+    } else if (config.ariaAutoLabel) {
```
rokotyan, line 74: I find this naming a little bit confusing: `config.ariaAutoLabel`. I think I would prefer `config.autoAriaLabel`. Or `config.autoDescription` and `config.description` instead of `config.ariaLabel`.

packages/ts/src/core/container/index.ts
```diff
@@ -62,15 +68,24 @@ export class ContainerCore {
...
+      description = this._getAriaDescription().trim()
+    }
+    if (description) {
```
rokotyan, line 77: If `config.ariaLabel` and `config.ariaAutoLabel` were set to `false` in a config update, the chart will still keep its old description.

And I would like to make this block look a little cleaner by moving the `description` logic inside `.attr('aria-label', ...)` or into a separate method.

```ts
    let description = ''
    if (config.ariaLabel) {
      description = config.ariaLabel
    } else if (config.ariaAutoLabel) {
      description = this._getAriaDescription().trim()
    }

    if (description) {
      select(this._container)
        .attr('aria-label', description)
        .attr('tabindex', '0')
```

packages/ts/src/core/component/index.ts
```diff
@@ -151,4 +151,9 @@ export class ComponentCore<
   public isDestroyed (): boolean {
     return !this.element
   }
+
+  // base class abstract method
+  public getAriaDescription (tickFormat: any): string {
```
rokotyan, line 156: We shouldn't use `any`

lee00678:
Closing this PR for now. Created an issue in the `issue-only` repo. We will reconsider the implementation in the future.

Issue #264, opened 2023-08-21
Accessibility: Automatic chart description generation


rokotyan:
TBA

curran:
LLMs might be really good at this. Here's example output from ChatGPT where I just pasted this image and said "Please describe this chart". Is this sort of text what you had in mind for this feature? It would be useful to have some examples of "good" chart descriptions for a handful of example charts.

![image](https://github.com/user-attachments/assets/32fac7c7-e249-4732-9d71-b8b6d996ea09)

This chart depicts U.S. recorded music revenues by format from around 1973 to 2019. The formats represented are:

- Vinyl (blue)
- Cassette (red)
- CD (orange)
- Download (green)
- Streaming (aqua)

The x-axis represents the years from approximately 1973 to 2019. The y-axis represents the revenue in a stacked bar format, with different colors indicating the contribution of each format to the total revenue for each year.

Key observations:
- Vinyl (blue) was a significant source of revenue in the early years but declined steadily after the early 1980s.
- Cassette (red) saw growth from the mid-1970s, peaking in the 1980s, and then declining through the 1990s.
- CD (orange) began to rise in the mid-1980s, becoming the dominant format through the 1990s and early 2000s before declining sharply after the mid-2000s.
- Downloads (green) started contributing around the early 2000s, showing significant growth until around 2013, after which it began to decline.
- Streaming (aqua) appeared in the mid-2000s and has been growing rapidly, becoming the largest source of revenue in the most recent years shown.

The chart highlights the shifting landscape of music consumption over time, from physical formats like vinyl, cassette, and CD, to digital formats such as downloads and streaming.

rokotyan:
@curran Thanks for sharing your thoughts, I love the idea and have actually thought about it too. I'm not sure how scalable it is for a library though (considering that we need to make users add their LLM API keys to use the feature).

I would like to see a dedicated package that could do this for any graphics on the webpage. May be a good idea for a new open source project :)

PR #265, opened 2023-08-21
Core | Container: Refactoring and pre-render step

rokotyan:
This PR introduces the preRender step that should contain logic required for rendering, like calculating scales, setting component sizes, etc ...

Additionally, I've added `public`, `private` and `protected` keywords to the container methods.

#264


rokotyan:
@reb-dev Can you please review this PR. This separation will make it easier to integrate the accessibility features we're currently working on

reb-dev (approved):

PR #268, opened 2023-08-25
Component | Axis: Setting label and tick text color via config; Dual-axis chart dev example

rokotyan:
This PR will allow users to configure axis label and tick text color via the config object. Also adding a dev example of a dual-axis chart.

image
SCR-20230825-kzmf
SCR-20230825-kzon


reb-dev (approved):
Nice!

Issue #152, opened 2023-03-01
What about Vue 3 support?


eladcandroid:

rokotyan:
There's no official Vue integration as of now. However you can still use Unovis with Vue because the core of the library is written in TypeScript.

We're accepting contributions, if you want to help us create a Vue integration, let us know!

errantmind:
I'm using Unovis with Vue3 and plain javascript and it works fine. I mostly referenced the Typescript docs and then mentally translated that to javascript where necessary. There were a few gotchas to get it to re-render properly when Vue state changes, but it is doable. Let me know if you run into any issues here @eladcandroid  and I'll try to help.

zernonia:
Hey @rokotyan ! Im the maintainer for [Shadcn Vue](https://github.com/radix-vue/shadcn-vue), and I would love to contribute to this Vue3 integration. Could you assign this to me? 😁

rokotyan:
@zernonia That's awesome! Can you send me a note to nikita@f5.com? I would like to discuss the process with you over a call if you're open to that.

rokotyan:
The Vue 3 integration has been merged into the main branch 🥳 Kudos to @zernonia for this amazing contribution!

It'll be officially announced with the 1.3.0 release, but you you want to use it already, use version `1.3.0-beta.3`: https://www.npmjs.com/package/@unovis/vue/v/1.3.0-beta.3

PR #271, opened 2023-09-07
Refactoring: New package for shared code

rokotyan:
Moving common framework integrations related code to `packages/shared`

#152


packages/angular/src/components/axis/axis.component.ts
```diff
@@ -191,8 +160,8 @@ export class VisAxisComponent<Datum> implements AxisConfigInterface<Datum>, Afte
   }
 
   private getConfig (): AxisConfigInterface<Datum> {
-    const { duration, events, attributes, x, y, id, color, xScale, yScale, excludeFromDomainCalculation, position, type, fullSize, label, labelFontSize, labelMargin, labelColor, gridLine, tickLine, domainLine, minMaxTicksOnly, tickFormat, tickValues, numTicks, tickTextFitMode, tickTextWidth, tickTextSeparator, tickTextForceWordBreak, tickTextTrimType, tickTextFontSize, tickTextAlign, tickTextColor, tickPadding } = this
```
reb-dev, line 194: @rokotyan Why are the properties inherited from `XYComponentConfigInterface` being excluded if we are keeping the ones from `ComponentConfigInterface`?

reb-dev (commented):
@rokotyan Thanks for doing this!

Left one comment above. In the angular wrapper, XY components that extend `Partial<XYComponentConfigInterface>` include some inherited props but not all. For consistency's sake we should probably include all or none right?

reb-dev (approved):

packages/angular/src/components/axis/axis.component.ts
rokotyan: That only affects  some of the components (Axis, Tooltip, Crosshair, Brush, FreeBrush). Those properties are technically useless for auxiliary components.

PR #275, opened 2023-09-13
Theme | New theme: Patterns

reb-dev:
The PR adds a new theme to the library.

#### Pattern Theme
When `document.body` has the class `theme-patterns` we automatically apply patterns of two types:
- **Fill patterns** for solid shapes (most cases). Users can customize fill patterns by assigning `--vis-pattern${i}` to a valid SVG def of a `mask` element. The default palette looks like:
Screen Shot 2023-10-03 at 2 56 27 PM

with the corresponding css variables:

```
--vis-pattern-fill0: var(--vis-pattern-fill-stripes-diagonal);
--vis-pattern-fill1: var(--vis-pattern-fill-dots);
--vis-pattern-fill2: var(--vis-pattern-fill-stripes-vertical);
--vis-pattern-fill3: var(--vis-pattern-fill-crosshatch);
--vis-pattern-fill4: var(--vis-pattern-fill-waves);
--vis-pattern-fill5: var(--vis-pattern-fill-circles);
```

- **Line patterns** for the _Line_ component and when _BulletLegend_'s `bulletShape` property is set to `"line"`. Users can customize these patterns by assigning `--vis-pattern-marker${i}` to a valid SVG def of a `marker` element or `--vis-pattern-dasharray${i}` to a valid value for the [stroke-dasharray](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-dasharray) property. The default palette looks like:
  
Screen Shot 2023-10-03 at 2 56 48 PM

with the corresponding css variables:

```
--vis-pattern-marker0: var(--vis-pattern-marker-circle);

--vis-pattern-marker1: var(--vis-pattern-marker-triangle);
--vis-pattern-dasharray1: 9 1;

--vis-pattern-marker2: var(--vis-pattern-marker-diamond);
--vis-pattern-dasharray2: 2;

--vis-pattern-marker3: var(--vis-pattern-marker-arrow);
--vis-pattern-dasharray3: 2 3 8 3;

--vis-pattern-marker4: var(--vis-pattern-marker-square);
--vis-pattern-dasharray4: 6;

--vis-pattern-marker5: var(--vis-pattern-marker-star);
--vis-pattern-dasharray5: 1 6;
```

Other changes:
- Added a theme selector to the dev demo app (toggle switch for dark/light theme, wand button toggles pattern theme)
- Tweaked Bullet Legend's viewBox sizing method to accommodate the patterns (tweaking the viewBox height instead of the width gives the best results for both types)

### DIFF
--- a/packages/ts/src/containers/single-container/index.ts
+++ b/packages/ts/src/containers/single-container/index.ts
@@ -57,6 +57,9 @@ export class SingleContainer<Data> extends ContainerCore {
       tooltip.setComponents([this.component])
     }
 
+    // Defs
+    this.element.appendChild(this._svgDefs.node())
+
     if (!preventRender) this.render()
   }
 
--- a/packages/ts/src/core/container/index.ts
+++ b/packages/ts/src/core/container/index.ts
@@ -20,6 +20,7 @@ export class ContainerCore {
   protected _requestedAnimationFrame: number
   protected _isFirstRender = true
   protected _resizeObserver: ResizeObserver | undefined
+  protected _svgDefs: Selection<SVGDefsElement, unknown, null, undefined>
   private _containerSize: { width: number; height: number }
 
   // eslint-disable-next-line @typescript-eslint/naming-convention
@@ -42,6 +43,7 @@ export class ContainerCore {
       .attr('height', ContainerCore.DEFAULT_CONTAINER_HEIGHT) // Overriding default SVG height of 150
       .attr('aria-hidden', true)
 
+    this._svgDefs = this.svg.append('defs')
     this.element = this.svg.node()
   }
 
--- a/packages/dev/src/App.module.css
+++ b/packages/dev/src/App.module.css
@@ -1,5 +1,5 @@
 @import url('https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600;700;900&display=swap');
-@import url('https://stackpath.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css');
+@import url('https://cdnjs.cloudflare.com/ajax/libs/font-awesome/6.4.2/css/all.min.css');
 
 * {
   box-sizing: border-box;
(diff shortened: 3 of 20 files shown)


packages/ts/src/containers/single-container/index.ts
```diff
@@ -57,6 +57,9 @@ export class SingleContainer<Data> extends ContainerCore {
       tooltip.setComponents([this.component])
     }
 
+    // Defs
+    this.element.appendChild(this._svgDefs.node())
```
rokotyan, line 61: Won't this append the defs over and over again on every `updateContainer` call?

packages/ts/src/core/container/index.ts
```diff
@@ -42,6 +44,13 @@ export class ContainerCore {
       .attr('height', ContainerCore.DEFAULT_CONTAINER_HEIGHT) // Overriding default SVG height of 150
       .attr('aria-hidden', true)
 
+    this._svgDefs = this.svg.append('defs')
+
+    if (select(`defs.${defsLabel}`).empty()) {
```
rokotyan, line 49: I find this line with the global `select` a bit confusing. I guess it's to prevent duplication of defs when there's more than 1 chart on the page. If so, it would benefit from having a comment with description.

Also it can be error prone: if a component containing the defs is removed from the page (e.g. user collapsed some view), other charts won't be able to find the defs anymore. Maybe it's worth moving this logic to `render`

rokotyan (commented):
Overall looks great, the implementation is clean and clear. I've only left a few comments.

packages/ts/src/containers/single-container/index.ts
reb-dev: No, because `removeAllChildren` is called initially (on [line 48](https://github.com/f5/unovis/blob/48662c5466d7ff09b1546ddabb9c519389c9c9f8/packages/ts/src/containers/single-container/index.ts#L48C5-L48C30))

rokotyan (approved):

Issue #236, opened 2023-06-19
Svelte v4 Support


pheuter:
[Svelte v4](https://github.com/sveltejs/svelte/releases/tag/svelte%404.0.0-next.0) is coming, would be nice to be able continue using `@unovis/svelte`.

rokotyan:
Thanks for the heads up @pheuter!

We've added Svelte 4 to the peer dependencies list in #237 and I plan to release it along with Unovis v1.2 within the next few days. The Svelte 4 integration hasn't been tested yet, so we'd appreciate it if you could report any problems you encounter.

rokotyan:
Unovis 1.2 with preliminary Svelte 4 support just got released

Issue #230, opened 2023-06-12
Website | Gallery: Fix sizing for Nested Donut example


reb-dev:
https://github.com/f5/unovis/assets/52078477/c884f3c9-67ff-4cd9-b601-566885722cd9

PR #278, opened 2023-09-21
Angular: Moving Gallery examples to `packages/angular`

ivegotwings:
Known issues: 
- Imports from `@unovis/ts/maps` don't work


packages/angular/angular.json
```diff
@@ -26,6 +26,103 @@
...
+          "skipTests": true
+        },
+        "@schematics/angular:service": {
+          "skipTests": true
+        },
+        "@schematics/angular:application": {
+          "strict": true
+        }
+      },
+      "root": "angulargalleryview",
```
rokotyan, line 61: The name `angulargalleryview` sounds overly complicated in my opinion. What if we just call it `gallery`?

packages/angular/angulargalleryview/src/main.ts
```diff
@@ -1,55 +1,53 @@
...
-
 import { NgModule, Component } from '@angular/core'
 import { platformBrowserDynamic } from '@angular/platform-browser-dynamic'
 import { BrowserModule } from '@angular/platform-browser'
 
 // Area
-import { NonStackedAreaChartModule } from './examples/non-stacked-area-chart/non-stacked-area-chart.module'
-import { StackedAreaModule } from './examples/stacked-area-chart/stacked-area-chart.module'
-import { StepAreaChartModule } from './examples/step-area-chart/step-area-chart.module'
+import { NonStackedAreaChartModule } from '../../../shared/examples/non-stacked-area-chart/non-stacked-area-chart.module'
```
rokotyan, line 7: Can we import examples from `@unovis/shared`, like other packages do?

packages/angular/package.json
```diff
@@ -43,10 +44,12 @@
     "tslib": "^2.3.1"
   },
   "peerDependencies": {
+    "@unovis/shared": "^1.2.1",
```
rokotyan, line 47: Since  `@unovis/shared` is not a public package I would change it to ` "@unovis/shared": "*"`. It'll make library versioning easier, otherwise we'll have to update our versioning scripts.

packages/angular/package.json
```diff
@@ -29,7 +29,8 @@
     "build": "npx ng build",
     "watch": "npx ng build --watch --configuration development",
     "test": "npx ng test",
-    "generate": "rollup -c; node .autogen.cjs; rm .autogen.cjs"
+    "generate": "rollup -c; node .autogen.cjs; rm .autogen.cjs",
+    "gallery:angular": "ng serve angulargalleryview"
```
rokotyan, line 33: Since we're already inside the `Angular` package, I don't think there's a need for the `:angular` suffix here.

rokotyan (changes requested):
Thanks @ivegotwings!

I've left a few comments to the code. I'm also getting a bunch of errors when I'm trying to run the gallery 
image
Can you please try doing a clean install `npm run install:clean` to make sure it works on your end?

We're also planning to adopt the following commit messaging rules. 
>
>`<Scope>( | <Sub-Scope>( | <Sub-Scope>)): Description( <#issue>)`
>
>Scope can be:
>• React, Angular, Vue, Svelte, Website, Dev, Shared
>• Core, Component, Container, Utils, Styles
>• Release
>• CI/CD
>• Misc
>
>Examples:
>• Component | Chord Diagram: Make link path generators dependent on `radiusScale` #228
>• Core | Container: Adding `_preRender` step; Refactoring #264
>• Component | NestedDonut: Making segment label font size configurable #230
>• Angular: Updating the wrappers #0
>• Website | Docs | LeafletMap: Adding information about `pointLabelColor` and `clusterLabelColor` #155
>• Svelte: Adding Svelte 4 to the supported `peerDependencies` list #236
>• Dev | New Example: Dual-axis chart [#261](https://github.com/f5/unovis/issues/261)
>• Release: 1.2.1

Can you please update your commit with a new name according to them?

packages/angular/package.json
```diff
@@ -43,10 +44,12 @@
     "tslib": "^2.3.1"
   },
   "peerDependencies": {
+    "@unovis/shared": "*",
+    "@unovis/ts": "^1.2.1",
```
rokotyan, line 48: @ivegotwings I think we need to specify the exact version here because we can't really guarantee the support of newer versions.

rokotyan (approved):

reb-dev (approved):

PR #279, opened 2023-09-22
Refactoring: From Config Classes to Objects

rokotyan:
Currently Unovis initializes config classes based on the corresponding TypeScript interfaces. There's a big problem with this approach: the default parameters in the class don't automatically inherit types from the interface, making our code more error-prone.

This PR moves away from using classes to using simple objects for component and container configuration. It solves the problem with the types, and it has already pointed out some of the existing bugs in the code.

The main downside of this approach is that it may be a breaking change on the TypeScript side of things, causing TypeScript errors in our users' code.


packages/ts/src/components/stacked-bar/config.ts
```diff
@@ -29,14 +31,18 @@ export interface StackedBarConfigInterface<Datum> extends XYComponentConfigInter
...
-  barPadding = 0.0
-  roundedCorners = 2
-  cursor = null
-  barMinHeight1Px = false
-  barMinHeightZeroValue = null
-  orientation = Orientation.Vertical
+export const StackedBarDefaultConfig: StackedBarConfigInterface<unknown> = {
+  ...XYComponentDefaultConfig,
+  // eslint-disable-next-line dot-notation
+  color: (d: unknown): string => d['color'],
```
reb-dev, line 37: See above

reb-dev (commented):
@rokotyan The change makes sense, I'm on board with it.

Besides the comments above,
- Any reason we should keep all of the `undefined` values in the default configs? It seems like a good chance to clean up the code a little.
- Stylistically - why not use camel case for the configs. i.e. `areaDefaultConfig` instead of `AreaDefaultConfig`? Since they are constants now not classes

reb-dev (approved):

reb-dev (approved):

PR #290, opened 2023-10-17
Refactoring: No Implicit Any

rokotyan:
Enabling `noImplicitAny` checks in tsconfig. Overall that should improve the code quality by allowing to catch potential errors while writing code.


rokotyan:
@reb-dev Can you please check the following TS errors in Nested Donut?
SCR-20231018-mtrv

My initial thought was that we need to pass an array to `HierarchyRectangularNode` but that generates even more errors.
SCR-20231018-mtnh

reb-dev:
@rokotyan Yeah so this is due to a discrepancy between the original data type (returned from d3's `hierarchy` layout function), and the type we use for Nested Donut (`NestedDonutSegment<Datum>`). The change occurs after the call to `partitionData.each()`. Basically,

Before:
```
typeof node == HierarchyRectangularNode<InternMap<string, number[]>>
typeof node.data == [
    key: string,
    values: InternMap<[string, number[]]> | number[]
]
```

After:
```
typeof node == HierarchyRectangularNode<NestedDonutSegmentDatum<Datum>>
typeof node.data == NestedDonutSegmentDatum<Datum>
```

So no need to change any existing types, it is the logic inside of function `_getHierarchyData` that needs to be reworked. Maybe creating a separate array of type `NestedDonutSegment<Datum>>[]` and adding the new nodes to it instead of modifying them in place would help?

rokotyan:
@reb-dev Thanks for the explanation, I understand the problem.

Will you be able to look into it? I've tried to quickly understand data transformations that are happening there but got stuck and didn't want to spend too much time on this. We can probably override the type of `node` inside the `each` loop to the correct one like this:
```ts
partitionData.each(node => {
  const n = node as unknown as ...
  node.data = {
    key: n.data[0],
    ...
```

And I would also leave a comment in the code that we're mutating the original data.

reb-dev:
@rokotyan I've added the changes (and added some comments as well) in a separate [PR](#296)  so I didn't block myself from merging this one. Do you want to merge it and rebase then I'll approve?

rokotyan:
@reb-dev Yeah, let's do it, thanks! Merging now.

rokotyan:
@reb-dev Actually, you'll need to merge it :)

reb-dev (approved):

Issue #286, opened 2023-10-10
Hidden tooltip overflows viewport


Coronon:
I am currently implementing quite a long `VisTooltip` on a `VisXYContainer`. I wrap the content of this tooltip using CSS, which works as expected. Sadly, when the tooltip was first shown on a larger screen and the viewport is resized to be smaller, the CSS `left` property on the tooltip container does not update which causes the viewport to overflow horizontally. As soon as the user hovers my chart again and the `left` value is recomputed+updated, this overflow goes away.

As a quick fix I propose to use `display: none` to hide the tooltip when it is not shown instead of `opacity: 0`. Another solution would be allowing people to overwrite the styles of the container itself instead of only it's content.

Version: `1.2.1(@unovis/ts@1.2.1)(svelte@4.2.1)`

rokotyan:
@Coronon Nice catch, it was bothering me as well.

@reb-dev Can you please look into this? If we simply use `display: none` we won't have smooth fade outs, but maybe we can come up with some trick that will make it work.

rokotyan:
@Coronon We've just released version `1.2.3` containing this fix

PR #292, opened 2023-10-18
Component | Tooltip | Fix: prevent container overflow when viewport size is reduced

reb-dev:
closes: #286

### DIFF
--- a/packages/dev/src/examples/auxiliary/tooltip/tooltip-resize-behavior/index.tsx
+++ b/packages/dev/src/examples/auxiliary/tooltip/tooltip-resize-behavior/index.tsx
@@ -1,0 +1,43 @@
+import React from 'react'
+import { Position, Scatter } from '@unovis/ts'
+import { VisXYContainer, VisAxis, VisScatter, VisTooltip } from '@unovis/react'
+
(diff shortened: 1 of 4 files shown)


reb-dev:
@rokotyan There is a small issue still. Even though the overflow is gone, the tooltip's position still takes multiple triggers to update. See the video below:

https://github.com/f5/unovis/assets/52078477/639574cb-ef96-4126-a144-8dce1c5fc27b

After investigating I discovered it is directly related to using `scrollWidth`. My original fix was to just use `clientWidth` instead, but then then breaks behavior for scrolling containers. For whatever reason though height isn't effected in the same way; from my testing `clientHeight === scrollHeight` but `clientWidth <= scrollWidth` even when they are under the same constraints.

Since it isn't a result of my changes here, I figured we can sill merge this and I can try to come up with a solution for the above issue in a separate PR.

rokotyan:
@reb-dev Thanks! 

Yes, the previous code had the same issue. Let's merge this PR and fix the remaining issue in the next one.

rokotyan (approved):

Coronon:
Thank you very much everyone for this quick fix 🎉😇

Issue #291, opened 2023-10-18
@unovis/svelte VisSingleContainer Props have no effect


kjell0w:
Hey,

when rendering a Donut Chart inside of a VisSingleContainer, the width, height, padding and margin properties have no effect on the Donut Chart. I also think the className prop does nothing, too. I would also suggest to rename it to just class for the Svelte Version, since className comes from JSX.

Heres a simple reproduction in Stackblitz: https://stackblitz.com/edit/zvpkrs?file=src%2FApp.svelte

PR #294, opened 2023-10-24
@unovis/svelte: Fixing SingleContainer props and adding support for `class` property

reb-dev:
closes: #291 

Previously, `$$restProps` object was ignored in the `initChart` call, resulting in missing props. I fixed the logic and did some minor refactoring to ensure update calls weren't being called unnecessarily.

Additionally, users can now add the `class` property for containers.

Note: Due to the way [svelte component styles](https://svelte.dev/docs/svelte-components#style) are scoped, the `:global` modifier is required for it to take effect. It is recommended to combine local and global to avoid conflicts and keep the scope as limited as possible. For example:

```html
<div>
    <VisSingleContainer class='chart'>...</VisSingleContainer>
</div>
<style>
    div :global(.chart) {
        /* Applies to elements in any component inside a <div>
        element belonging to this component. ✅ Recommended */
    }

    :global(.chart) {
        /* Applies to all elements in any component in the document
        where class="chart". ⛔ Works, but not recommended */
     }

    .chart {
        /* Strictly applies to elements (not components) that belong to this
        component. ❌ Doesn't work for `@unovis/svelte` containers */
    }
</style>
```

### DIFF
--- a/packages/svelte/.eslintrc.cjs
+++ b/packages/svelte/.eslintrc.cjs
@@ -21,6 +21,7 @@ module.exports = {
     rules: {
       'svelte/indent': ['error'],
       'no-use-before-define': 'off',
+      'no-undef-init': 'off',
       '@typescript-eslint/no-unsafe-call': 'off',
       '@typescript-eslint/no-unsafe-assignment': 'off',
       '@typescript-eslint/no-unsafe-member-access': 'off',
--- a/packages/svelte/src/containers/xy-container/xy-container.svelte
+++ b/packages/svelte/src/containers/xy-container/xy-container.svelte
@@ -8,7 +8,7 @@
   }
 
   // Props
-  export let data: Datum[]
+  export let data: Datum[] = undefined
   export let className = ''
   /**
    * CSS class string. Requires `:global` modifier to take effect. _i.e._
--- a/packages/svelte/src/containers/xy-container/xy-container.svelte
+++ b/packages/svelte/src/containers/xy-container/xy-container.svelte
@@ -3,8 +3,28 @@
   import { onMount, setContext } from 'svelte'
 
   type Datum = $$Generic
+  interface $$Props extends XYContainerConfigInterface<Datum> {
+    data?: Datum[];
+  }
 
+  // Props
   export let data: Datum[]
+  export let className = ''
+  /**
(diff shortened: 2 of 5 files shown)


packages/svelte/src/containers/xy-container/xy-container.svelte
```diff
@@ -5,6 +5,8 @@
   type Datum = $$Generic
 
   export let data: Datum[]
+  export let className = ''
```
rokotyan, line 8: @reb-dev Do you know if there's a way to add comments to svelte props an make them show up as you type your html? That would be great to add your note on using `:global` here to make it more obvious to the users.

reb-dev: @rokotyan Yes! I've just added JS docs in the last push. Couldn't get the same level as detail as above but I think it gets the point across anyway.

rokotyan: @reb-dev Awesome, thanks!

reb-dev: @rokotyan Nevermind, I guess it broke the build. Looking into alternatives

packages/svelte/.eslintrc.cjs
```diff
@@ -21,6 +21,7 @@ module.exports = {
     rules: {
       'svelte/indent': ['error'],
       'no-use-before-define': 'off',
+      'no-undef-init': 'off',
```
reb-dev, line 24: @rokotyan Context: The linter has been removing `= undefined` initializations in the pre-commit script. However for the sake of svelte components, it is necessary to include them because it signifies that a prop is optional (see https://svelte.dev/docs/svelte-components#script-1-export-creates-a-component-prop)
```
// Required prop
export let prop: PropType;

// Optional prop
export let prop: PropType = undefined
```

Previously the only consequence was misguided console warnings (for example _XYContainer is missing required prop `data`_) but now that I've added prop type defs for better IntelliSense, we need to include these initializations to prevent pipeline errors.

rokotyan: @reb-dev Nice catch!

rokotyan (approved):

PR #330, opened 2024-01-25
Core Component and Tooltip event handling tweaks

rokotyan:
This PR tweaks the behavior of the core component and the tooltip related to events:
* Core | Component | Events: Check if the callback exists before execution
* Component | Tooltip: Allowing `triggers` to be `undefined` or `null`
* Component | Tooltip: If the container is `body`, don't change its position style because it can have unwanted effect


reb-dev (approved):
Looks good.

rokotyan:
@reb-dev Great, feel free to merge when you're ready. I won't be able to do it myself.

PR #344, opened 2024-03-05
Unovis React: Fixing ref initialization

rokotyan:
Previously component refs were `undefined` due to the component initialization cycle. Changing the ref logic to use a getter to always return the latest instance of the component.


reb-dev (approved):
@rokotyan Nice! Thanks for the fix.

Issue #355, opened 2024-03-21
Crosshair position incorrectly with multiple Area


zernonia:
When using multiple `Area` with `Crosshair`, the positioning is wrong for 2nd Area onwards.
Expecting the crosshair to position correctly.

Reproduction link: https://stackblitz.com/edit/estk3b?file=src%2FApp.ts
Based upon the [Gallery Example](https://unovis.dev/gallery/view?collection=Lines%20and%20Areas&title=Non-Stacked%20Area%20Chart)

![image](https://github.com/f5/unovis/assets/59365435/fcb1306a-dd65-4c71-8fc0-c18182f0b9e2)

PR #356, opened 2024-03-21
fix: crosshair with multiple area issue

zernonia:
Fix https://github.com/f5/unovis/issues/355

`cla-signed` since on previous contribution 😁


rokotyan:
Thanks for fixing this @zernonia!

reb-dev (approved):

PR #370, opened 2024-04-16
Website | Gallery: Scatter Plot with Varied Shape

lee00678:
Adding an example to gallery for scatter plot with varied shape.
[issue](https://github.com/unovis/issues-only/issues/40)

I'm not sure if the best way to map `shapes` is to use an array inside data. If there are better ways, please let me know, I'm happy to change it.


packages/shared/examples/shaped-scatter-plot/shaped-scatter-plot.tsx
```diff
@@ -0,0 +1,46 @@
...
+      <VisXYContainer data={data} height={'60vh'} scaleByDomain={true}>
+        <VisScatter
+          x={useCallback((d: DataRecord) => +(new Date(d.date)), [])}
+          y={useCallback((d: DataRecord) => d.trainedParam, [])}
+          shape={useCallback((d: DataRecord) => shapeScale(sumCategories(d.owner)), [])}
+          size={15}
+          label={useCallback((d: DataRecord) => d.name, [])}
+          labelPosition={Position.Bottom}
+          labelHideOverlapping={true}
+          labelColor="grey"
```
rokotyan, line 35: The `grey` color looks a bit dark in the `dark` mode.

image

rokotyan (approved):
Love this new example! My only concern is the label color in the dark mode.

packages/shared/examples/shaped-scatter-plot/data.ts
```diff
@@ -0,0 +1,107 @@
...
+export const shapes = [
+  'circle',
+  'cross',
+  'diamond',
+  'square',
+  'star',
+  'triangle',
+]
+
+export const sumCategories = (category: string): string => {
```
reb-dev, line 27: I'm not sure this function is necessary. I see you're using it to map category to the color/shape, but `scaleOrdinal` should already implement that logic for you. See my other comment below

packages/shared/examples/shaped-scatter-plot/shaped-scatter-plot.component.html
```diff
@@ -0,0 +1,17 @@
...
+<vis-bullet-legend [items]="legendItems"></vis-bullet-legend>
+<vis-xy-container [data]="data" [height]="600">
+  <vis-scatter
+    [x]="getX"
+    [y]="getY"
+    [color]="getColor"
+    [size]="15"
+    [shape]="getShape"
+    [label]="getLabel"
+    [labelPosition]="'bottom'"
```
reb-dev, line 11: Not a big deal but for consistency's sake I would use this syntax: `labelPosition="bottom"`, like you have for the cursor property below it.

reb-dev (commented):
@lee00678 Great example, thanks for making it! Overall looks good but I left a couple comments.

packages/shared/examples/shaped-scatter-plot/data.ts
lee00678: `sumCategories` also put any owner that's not in `const categories` into `Other`.

packages/shared/examples/shaped-scatter-plot/shaped-scatter-plot.tsx
lee00678: Good call. Updated to use the default color.

packages/shared/examples/shaped-scatter-plot/shaped-scatter-plot.component.html
reb-dev: True! I would just drop that line entirely then.

And `cursor="pointer"` is fine

packages/shared/examples/shaped-scatter-plot/data.ts
reb-dev: Fair point. I think it still works without it but this way it clearer, so let's keep it

lee00678: Do you mean we can get rid of `sumCategories` entirely and call `getColor = (d: DataRecord): string => colorScale(d.owner)` and `shapeScale(d.owner)` directly? I'm not sure if that would work. Because then, in the data set, if the `owner` isn't part of the 6 categories, (like owner: 'Inflection'), it will get mapped to a different shape and color instead of `Other` right? Or am I missing something? 

I don't want to have inaccurate examples, so I'm happy to update if that's the case!

reb-dev (approved):

packages/shared/examples/shaped-scatter-plot/data.ts
reb-dev: @lee00678 Yeah so I actually just tried it myself and it does work (feel free to try it too and verify). I think under the hood the `scaleOrdinal` automatically bins extraneous values into the last value of the provided range.

But that doesn't mean your example is inaccurate, if anything it is more clear. So I approved this PR you can go ahead and merge if you like. Or if you want to change it, that's fine too. I'll leave it up to you :)

lee00678: Thanks for the review! I didn't know this feature about `scaleOrdinal`.

PR #371, opened 2024-04-16
Website | Gallery: Update Basic Grouped Bar Chart Legend

lee00678:
Quick update for the legend to only capitalize the first letter. ([Issue](https://github.com/unovis/issues-only/issues/39))

Please see screenshot:
Screenshot 2024-04-17 at 10 57 06 AM

### DIFF
--- a/packages/shared/examples/basic-grouped-bar/basic-grouped-bar.component.ts
+++ b/packages/shared/examples/basic-grouped-bar/basic-grouped-bar.component.ts
@@ -1,5 +1,5 @@
 import { Component } from '@angular/core'
-import { data, colors, ElectionDatum } from './data'
+import { data, colors, capitalize, ElectionDatum } from './data'
 
 @Component({
   selector: 'basic-grouped-bar',
@@ -17,7 +17,7 @@ export class BasicGroupedBarComponent {
   data = data
 
   legendItems = Object.entries(colors).map(([n, c]) => ({
-    name: n.toUpperCase(),
+    name: capitalize(n),
     color: c,
   }))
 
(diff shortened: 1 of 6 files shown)


rokotyan:
Thanks for fixing this @lee00678!

I've a small thought: since it is one of our official examples, I find it a little "dirty" to use `n.charAt(0).toUpperCase() + n.slice(1)` there. In the real word use case, I would define a function to do this, or used `text-transform: capitalize` in CSS. 

I don't think we want to overcomplicate this example by adding unnecessary CSS. We can probably add a "capitalize" function to `data.ts` or simply add a labels map there: `{ republican: 'Republican', ... }`. Let me know what you think.

P.S. We usually have a practice of adding screenshots and videos to our PRs to simplify the review process. That would be great if you could do it too.

lee00678:
> Thanks for fixing this @lee00678!
> 
> I've a small thought: since it is one of our official examples, I find it a little "dirty" to use `n.charAt(0).toUpperCase() + n.slice(1)` there. In the real word use case, I would define a function to do this, or used `text-transform: capitalize` in CSS.
> 
> I don't think we want to overcomplicate this example by adding unnecessary CSS. We can probably add a "capitalize" function to `data.ts` or simply add a labels map there: `{ republican: 'Republican', ... }`. Let me know what you think.

Agree with not adding CSS to complicate the example. I like the idea of adding a `capitalize` function in `data.ts`. I think it's a common use case and would be nice to provide an small example for that? I will update. Thanks for the feedback!
> 
> P.S. We usually have a practice of adding screenshots and videos to our PRs to simplify the review process. That would be great if you could do it too.

Gotcha.

rokotyan:
Yep, that works! Thanks for addressing the comments.

reb-dev (approved):
Looks good. Thanks @lee00678

PR #372, opened 2024-04-20
Shared | Autogen: Adding support for standalone components

reb-dev:
This PR updates component autogeneration logic so that all Unovis components (not containers) can be automatically generated when running `npm run generate`. Previously, our standalone components BulletLegend, LeafletMap and LeafletFlowMap had to be updated manually in the React and Angular wrappers.

In addition to autogen, another motivation for this change is to ensure our list in `shared/integrations/components` is complete. This allows the list be used for multiple purposes across the codebase (docs, commitlint, etc).

closes: unovis/issues-only#41

### DIFF
--- a/packages/react/autogen/component.ts
+++ b/packages/react/autogen/component.ts
@@ -5,61 +5,66 @@ export function getComponentCode (
   generics: GenericParameter[] | undefined,
   importStatements: { source: string; elements: string[] }[],
   dataType: string | null = 'any',
-  elementSuffix = 'component'
+  elementSuffix = 'component',
+  isStandAlone = false
 ): string {
   const genericsStr = generics ? `<${generics?.map(g => g.name).join(', ')}>` : ''
   const genericsDefStr = generics
     ? `<${generics?.map(g => g.name + (g.extends ? ` extends ${g.extends}` : '') + (g.default ? ` = ${g.default}` : '')).join(', ')}>`
     : ''
+  const componentType = `${componentName}${genericsStr}`
+  const refType = isStandAlone ? 'HTMLDivElement' : `VisComponentElement<${componentType}>`
+  const onDestroy = isStandAlone ? 'c?.destroy()' : `{
+    componentRef.current = undefined
+    c.destroy()
+  }`
   return `// !!! This code was automatically generated. You should not change it !!!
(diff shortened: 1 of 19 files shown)


packages/react/autogen/component.ts
```diff
@@ -5,61 +5,66 @@ export function getComponentCode (
...
-    const component = componentRef.current
+    ${isStandAlone ? '' : 'const component = componentRef.current'}
     ${dataType ? 'if (props.data) component?.setData(props.data)' : ''}
-    component?.setConfig(props)
+    component?.${componentName === 'BulletLegend' ? 'update ' : 'setConfig'}(props)
   })
 
-  useImperativeHandle(fRef, () => ({ get component () { return componentRef.current } }), [])
-  return <vis-${elementSuffix} ref={ref} />
+  useImperativeHandle(fRef, () => ({ get component () { return component${isStandAlone ? '' : 'Ref.current'} } }), [])
```
rokotyan, line 66: The `Ref.current`statement looks a bit suspicious, is it correct?

rokotyan (approved):
Thanks for implementing this @reb-dev!

I'm also planning to allow stand-alone components like Bullet Legend to be appended straight to the provided DOM node to simplify styling when used with UI frameworks like React. Currently we have double divs — one comes from the framework (e.g. via a ref), and another one is created by a component.

For example, here's a Bullet Legend change that I've already tested with a published beta:
```ts
constructor (element: HTMLElement, config?: BulletLegendConfigInterface) {
    this._container = element
    this.div = config?.renderIntoProvidedDomNode ? select(this._container) : select(this._container).append<HTMLElement>('div')
...
```

And the the React wrapper the initialization looks like this:
```ts
const c = new BulletLegend(container.current as HTMLDivElement, { ...props, renderIntoProvidedDomNode: true })
```

If I create a Pull Request with these changes, will you be able to incorporate them in this autogen update too?

packages/react/autogen/component.ts
reb-dev: Nah that is definitely a typo, thanks for catching it. Updated.

reb-dev:
> If I create a Pull Request with these changes, will you be able to incorporate them in this autogen update too?

@rokotyan Sure 👍

lee00678 (approved):

PR #374, opened 2024-04-23
Component | Bullet Legend: Vertical orientation, Safari rendering fixes, ...

rokotyan:
This PR implements a bunch of small tweaks to the Bullet Legend component add the orientation property.

Summary of changes:
- New `orientation` config property
- New `renderIntoProvidedDomNode` property. When set to `true` the legend items will be appended straight to the provided in the constructor DOM node. It is needed to allow for more flexible styling when the component used with the UI frameworks and they initialize their own `div` for the component. I say "provided DOM node" instead of just "container" to avoid confusion with our Container components.
- Appearance fixes for Safari
- More CSS variables to control how the legend looks
- New Dev example
- Updated Docs
- Updated the Bullet Legend React wrapper to accept a `className`

Here is an example of how the vertical orientation can be used:

https://github.com/f5/unovis/assets/755708/a5a65b8e-4df5-4336-a7ed-87c5a1d34b28


rokotyan:
@reb-dev Sorry, I didn't have time to implement the `renderIntoProvidedDomNode` config for the other components. Also I've only updated and tested the React wrapper.

rokotyan:

lee00678:
@rokotyan this is great! @reb-dev and I were talking about adding this. It's nice to see how your approach is.

reb-dev (commented):
@rokotyan Looks great! Thanks for doing this.

Quick question - if we always override `renderIntoProvidedDomView` in the wrapper components, then setting this property will have no effect unless using BulletLegend from `@unovis/ts`, right? If so, we should probably clarify that in the JSDoc

rokotyan:
@reb-dev Yes, you're right. To be honest, I don't see any use case when not overriding it would make sense (when using Unovis with a UI framework). And I think it's a good idea to add this information to JSDOC.

lee00678 (approved):
Looks good to me.

PR #390, opened 2024-05-17
Website | Gallery: Range plot

lee00678:
Added an example for range plot to the gallery as part of the composite chart section. I suppose this can be a separate component eventually. But since it can be done with the current components I figured it could be useful for people.

My approach was to use 2 scatter plots and a line chart. If there's a better approach I'm happy to update.
 
range-plot
range-plot-dark


packages/shared/examples/range-plot/range-plot.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { VisXYContainer, VisBulletLegend, VisTooltip, VisAxis, VisScatter, VisLine } from '@unovis/react'
+
+import { data, DataRecord, processLineData } from './data'
+
+const height = 1600
+const yScale = Scale.scalePoint([0, 800]).domain(data.map(d => d.occupation))
+const lineData = processLineData(data)
+
+export default function RangePlot (): JSX.Element {
+  const legendItems = [{ name: 'Women', color: '#FF6B7E' }, { name: 'Men', color: '#4D8CFD' }]
```
rokotyan, line 12: If we want to show how to use Unovis components efficiently with React, you should use `useMemo` or `useCallback` everywhere we initialize a new object, array, or function.

packages/shared/examples/range-plot/range-plot.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const style: React.CSSProperties = { width: '100%', height: height }
```
rokotyan, line 13: Here

packages/shared/examples/range-plot/range-plot.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const styleAb: React.CSSProperties = { position: 'absolute', top: 0, left: 0, width: '100%', height: height }
```
rokotyan, line 14: Here

packages/shared/examples/range-plot/range-plot.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const tooltipTriggers = {
```
rokotyan, line 15: Here

packages/shared/examples/range-plot/range-plot.tsx
```diff
@@ -0,0 +1,58 @@
...
+          data={data}
+          x={useCallback((d: DataRecord) => d.women, [])}
+          y={useCallback((d: DataRecord) => yScale(d.occupation), [])}
+          color={'#FF6B7E'}
+          size={10}
+          style={styleAb}
+        ></VisScatter>
+
+        <VisAxis type='x' numTicks={5} label={'Yearly Salary'} />
+        <VisAxis type='y' tickFormat={(_, i: number) => data[i].occupation} numTicks={data.length} gridLine={false} />
```
rokotyan, line 53: And here

rokotyan (approved):
Thanks @lee00678! Agreed that it looks a little bit like a hack but it's still good to have an example of what's possible with Unovis.

Left you a few React-related comments

packages/shared/examples/range-plot/range-plot.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { data, DataRecord, processLineData } from './data'
+
+const height = 1600
+const yScale = Scale.scalePoint([0, 800]).domain(data.map(d => d.occupation))
+const lineData = useMemo(() => processLineData(data), [])
+
+export default function RangePlot (): JSX.Element {
+  const legendItems = [{ name: 'Women', color: '#FF6B7E' }, { name: 'Men', color: '#4D8CFD' }]
+  const style: React.CSSProperties = useMemo(() => ({ width: '100%', height: height }), [])
+  const styleAb: React.CSSProperties = useMemo(() => ({ position: 'absolute', top: 0, left: 0, width: '100%', height: height }), [])
```
reb-dev, line 14: Correct me if I'm wrong, but I don't think we need this style declaration. The two scatters should render automatically without additional styling.

packages/shared/examples/range-plot/styles.css
reb-dev, line 1: Is this file used anywhere?

packages/shared/examples/range-plot/range-plot.tsx
```diff
@@ -0,0 +1,59 @@
...
+      Men average pay: $${Intl.NumberFormat().format(d.men)} </br>
+      Pay gap: $${Intl.NumberFormat().format(d.gap)}</br>
+    `,
+    }), [])
+
+  return (
+    <>
+      <h2>Gender Pay Gap</h2>
+      <VisBulletLegend items={legendItems}/>
+      <VisXYContainer style={style}>
```
reb-dev, line 28: I think we can just get away with `height={height}` here instead

packages/shared/examples/range-plot/range-plot.component.ts
```diff
@@ -0,0 +1,33 @@
...
+@Component({
+  selector: 'range-plot',
+  templateUrl: './range-plot.component.html',
+})
+export class RangePlotComponent {
+  height = 1600
+  data = data
+  lineData = processLineData(data)
+  legendItems = [{ name: 'Women', color: '#FF6B7E' }, { name: 'Men', color: '#4D8CFD' }]
+  style = { width: '100%', height: this.height }
```
reb-dev, line 15: This doesn't seem to be used in the template. I would just remove it since it is already declared there:
```
<vis-xy-container
  [width]="'100%'"
  [height]="height"
>
```

packages/shared/examples/range-plot/range-plot.svelte
```diff
@@ -0,0 +1,73 @@
...
+
+  import { VisXYContainer, VisBulletLegend, VisTooltip, VisAxis, VisScatter, VisLine } from '@unovis/svelte'
+  import { Scale, Scatter } from '@unovis/ts'
+  import { data, DataRecord, processLineData } from './data'
+
+  const height = 1600
+  const yScale = Scale.scalePoint([0, 800]).domain(data.map(d => d.occupation))
+  const lineData = processLineData(data)
+  const legendItems = [{ name: 'Women', color: '#FF6B7E' }, { name: 'Men', color: '#4D8CFD' }]
+  const style: React.CSSProperties = { width: '100%', height: height }
```
reb-dev, line 11: Same comment from the angular example above. I would remove in favor of template declaration

packages/shared/examples/range-plot/range-plot.ts
```diff
@@ -0,0 +1,68 @@
+import { XYContainer, Line, Scale, Scatter, Axis, BulletLegend, Tooltip } from '@unovis/ts'
+import { data, DataRecord, processLineData, LineDataRecord } from './data'
+
+const height = 1782
+const container = document.getElementById('vis-container')
+
+const legendItems = [{ name: 'Women', color: '#FF6B7E' }, { name: 'Men', color: '#4D8CFD' }]
+const style: React.CSSProperties = { width: '100%', height: height }
```
reb-dev, line 8: See above

packages/shared/examples/range-plot/range-plot.vue
```diff
@@ -0,0 +1,64 @@
+<script setup lang="ts">
+import { VisXYContainer, VisBulletLegend, VisTooltip, VisAxis, VisScatter, VisLine  } from '@unovis/vue'
+import { Scale, Scatter } from '@unovis/ts'
+import { data, DataRecord, processLineData } from './data'
+
+const height = 1600
+const yScale = Scale.scalePoint([0, 800]).domain(data.map(d => d.occupation))
+const lineData = processLineData(data)
+const legendItems = [{name: "Women", color: '#FF6B7E'}, {name: 'Men', color: '#4D8CFD'}]
+const style: React.CSSProperties = { width: '100%', height: height }
```
reb-dev, line 10: See above

reb-dev (commented):
I really like the look of this example!

My only qualm is in regard to the style declarations.
- The container width by default is 100%, I think we can just specify the height directly through the height property
- I don't think we need to make positions absolute or have additional stylings for the scatter components, they should look the same without them (please correct me if I'm mistaken or missing something)

Neither is a huge deal, but I feel like the less code we have in our examples, the better. So if we can get away with removing some excess variables that would be ideal.

packages/shared/examples/range-plot/range-plot.tsx
lee00678: You are absolutely right. Thanks for catching it.

reb-dev (approved):

PR #392, opened 2024-05-21
Component | Brush: Additional styling options via CSS variables

rokotyan:
### What's new:
- I've added new CSS variables and changed the way we define them
- New dev example
- Updated the docs using newly exported `cssVarDefaults`. This way we can automatically fill the content of the CSS Variables section in the docs.

image

image

image


lee00678 (approved):
Thanks for the updates! Looks good to me.

reb-dev (approved):

PR #394, opened 2024-05-28
Component | Axis: Add tick label rotation

lee00678:
Following this [PR](https://github.com/f5/unovis/pull/315). Moved `rotate` out of `FitMode`.

- [x] Adding a new variable called `tickTextAngle` to allow user to pass in a number for label rotation. This feature should be used along with `tickTextAlign` to get the desired result (we want to provide more control to the user). 
- [x] New dev example
- [x] Add related documentation  

Screenshot 2024-05-28 at 2 31 05 PM

### DIFF
--- a/packages/dev/src/examples/xy-components/axis/axis-rotation/index.tsx
+++ b/packages/dev/src/examples/xy-components/axis/axis-rotation/index.tsx
@@ -1,0 +1,73 @@
+import React, { useRef } from 'react'
+import { VisXYContainer, VisAxis, VisLine, VisGroupedBar } from '@unovis/react'
+import { XYDataRecord, generateXYDataRecords } from '@src/utils/data'
+import { Scale } from '@unovis/ts'
+
+export const title = 'Axis with Ticks Rotation'
+export const subTitle = 'Generated Data'
+
+export const component = (): JSX.Element => {
+  const accessors = [
+    (d: XYDataRecord) => d.y,
+    (d: XYDataRecord) => d.y1,
+    (d: XYDataRecord) => d.y2,
+    () => Math.random(),
+    () => Math.random(),
+  ]
+  return (
+    <>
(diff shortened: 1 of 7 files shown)


reb-dev (commented):
@lee00678 , looks good!

I noticed an edge case- when a label is very long at the end, the label gets cut off on initial render. On resize, it updates correctly but if we refresh the page we see it again.

https://github.com/f5/unovis/assets/52078477/a05ae7d6-a851-4567-a41e-3ab53d4b2b43

lee00678:
> @lee00678 , looks good!
> 
> I noticed an edge case- when a label is very long at the end, the label gets cut off on initial render. On resize, it updates correctly but if we refresh the page we see it again.
> 
I think this may be an existing issue. Perhaps we should address this in a separate PR?
(Screenshot taken from`main`)

https://github.com/f5/unovis/assets/5026041/c47974b8-e3e6-4938-87c3-bc95c8480c33

reb-dev:
@lee00678
> I think this may be an existing issue. Perhaps we should address this in a separate PR? (Screenshot taken from`main`)

You're totally right. Let's move to a separate issue.

rokotyan:
@reb-dev @lee00678 Confirming that the issue has been there before. It's a bit tricky to calculate the required margins automatically and apparently the current logic has some flaws. Let me know if you decide to take a look into it, I can give you a quick walk through explaining how it works.

lee00678:
> @reb-dev @lee00678 Confirming that the issue has been there before. It's a bit tricky to calculate the required margins automatically and apparently the current logic has some flaws. Let me know if you decide to take a look into it, I can give you a quick walk through explaining how it works.

Yea margin calculation could be tricky. I took a brief look at it while implementing this feature. I will add an item to our issue list so we don't forget about it.

reb-dev (approved):

Issue #393, opened 2024-05-24
Force graph doesn't have force physics after rendering


mossy426:
This might be a misunderstanding, but we have properties for "force" in the layout settings like so:

```
type GraphForceLayoutSettings = {
  /** Preferred Link Distance. Default: `60` */
  linkDistance?: number;
  /** Link Strength [0:1]. Default: `0.45` */
  linkStrength?: number;
  /** Charge Force (<0 repulsion, >0 attraction). Default: `-500` */
  charge?: number;
  /** X-centring force. Default: `0.15` */
  forceXStrength?: number;
  /** Y-centring force. Default: `0.25` */
  forceYStrength?: number;
}
```

And we also have the "fixNodePositionAfterSimulation" property within that as well.

It doesn't seem to matter if you set fixNodePositionAfterSimulation to true or false.  Even in the gallery you guys have, the nodes are always fixed.  It seems the only force for a force graph is upon initialization of the graph, otherwise everything is pinned and not acted on by forces.

In our Microbetrace application, you can change the charge, force strength, etc. of the network at any time.  Is this not possible in Unovis?/Are all nodes pinned after instantiation regardless of whatever layout you choose?  

Might be more of a discussion than an issue, but please let me know :)

rokotyan:
@mossy426 Can you please tell which version of Unovis do you use?

We had a bug in the code that has been fixed in one of the latest betas but didn't make it to the main release yet.

mossy426:
All good!  I'm using the 1.4.1 build that was suggested.

rokotyan:
@mossy426 It turned out that `1.4.1` doesn't contain the fix that was published in `1.4.1-alpha.8`. We'll be releasing 1.4.2 soon. In the meantime, can you please check if it work on `1.4.1-alpha.8` for you?

mossy426:
@rokotyan I updated the environment on stackblitz for a quick test to use 1.4.1-alpha.8 for both packages and it appears to still not work, even with fixNodePositionAfterSimulation set to false

rokotyan:
@mossy426 Thank @mossy426! I think I know what was causing this. I've just published new package versions `1.4.2-alpha.2` to test. They should fix the problem

rokotyan:
@mossy426 Strange, NPM shows it's there. Let's wait a bit and I'll try to release another version if it still doesn't work.

[...]

PR #396, opened 2024-05-31
Component | Graph: Layout recalculation and persistency fix

rokotyan:
This PR fixes layout problems reported in #393:
- `fixNodePositionAfterSimulation` not working correctly in Force Layout
- Layout update when dynamically switching between different layouts

### DIFF
--- a/packages/dev/src/examples/networks-and-flows/graph/graph-dynamic-layout/index.tsx
+++ b/packages/dev/src/examples/networks-and-flows/graph/graph-dynamic-layout/index.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useState } from 'react'
+import React, { useEffect, useMemo, useState } from 'react'
 import { VisSingleContainer, VisGraph } from '@unovis/react'
 import { GraphLayoutType } from '@unovis/ts'
 import { generateNodeLinkData } from '@src/utils/data'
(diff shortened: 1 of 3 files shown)


reb-dev (commented):
@rokotyan I made a demo example to test this, I'm getting strange behavior if any panning/zooming occurs. Should we reset the transform on layout change as well?

https://github.com/f5/unovis/assets/52078477/837f3859-9b98-4335-a71c-40b6c0d22241

lee00678:
@rokotyan I tried to modify one of the gallery example to test this. It seems once the layout changed, the chart no longer fit in the view window. 

https://github.com/f5/unovis/assets/5026041/ed9996d8-4f02-4e21-b32b-452c319b1a6b

rokotyan:
@reb-dev @lee00678 Thanks for testing this! I didn't see such behavior in my environment, let me take another look.

rokotyan:
@reb-dev Can you please push your example to the branch?

rokotyan:
Thanks @reb-dev

My latest commits should take care of that and also fix https://github.com/f5/unovis/issues/393

reb-dev:
@rokotyan Testing it again I'm discovering inconsistent behavior with `fixNodePositionAfterSimulation`.

Currently if you dynamically add data to the graph when it is set to `true`, the layout doesn't rearrange. When `false`, it does. However if you switch to `force` layout type from a different layout with `fixNodePositionAfterSimulation` set to `true`, the layout still rearranges on data updates.

Also could you clarify what the expected behavior is for this property?

rokotyan:
@reb-dev You're right, something wrong is happening when changing the layout. Thanks for catching this, I'll investigate

rokotyan:
@reb-dev Found the problem and fixed, see the last two commits. Thanks again for testing this thoroughly.

reb-dev (approved):
Thanks for the updates @rokotyan, looks good to me!

PR #397, opened 2024-06-03
Component | Graph: Allow user to set node positions

lee00678:
- Simply extend `GraphInputNode` to allow the user to set node positions. This should make save and reload chart possible. 
- Added a dev example for this
- Updated Doc

### DIFF
--- a/packages/dev/src/examples/networks-and-flows/graph/graph-node-positions/index.tsx
+++ b/packages/dev/src/examples/networks-and-flows/graph/graph-node-positions/index.tsx
@@ -1,0 +1,24 @@
+import React, { useCallback, useMemo, useRef } from 'react'
+import { VisSingleContainer, VisGraph, VisGraphRef } from '@unovis/react'
+import { generatePrecalculatedNodeLinkData, NodeDatum, LinkDatum } from '@src/utils/data'
+
+export const title = 'Graph: Node Positions'
+export const subTitle = 'Pass in node positions'
+
+export const component = (): JSX.Element => {
+  const data = useMemo(() => generatePrecalculatedNodeLinkData(), [])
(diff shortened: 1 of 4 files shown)


reb-dev:
@lee00678 I don't see a new dev example, did you forget to add it in in this commit https://github.com/f5/unovis/pull/397/commits/6786b79b9e14caa570620f144f0c68da06fb89f9?

reb-dev (commented):
I think we need to add some more implementation for this.

Right now, providing x and y will automatically force the node positions (at least for `GraphLayoutType.Force`, the other ones would get overridden I think). Shouldn't we prevent that from happening if `GraphLayoutType.Precalculated` isn't specified?

lee00678:
> I think we need to add some more implementation for this.
> 
> Right now, providing x and y will automatically force the node positions (at least for `GraphLayoutType.Force`, the other ones would get overridden I think). Shouldn't we prevent that from happening if `GraphLayoutType.Precalculated` isn't specified?

That's a good point. I think you are right. I will update.

reb-dev:
My [previous comment](https://github.com/f5/unovis/pull/397#pullrequestreview-2097063339) no longer applies after merging https://github.com/f5/unovis/pull/396, we can just add `x` and `y` to nodes without it interfering with force layout.

@lee00678 Can you revert this PR to include your original changes? Then we can merge it

lee00678:
> @lee00678 Can you revert this PR to include your original changes? Then we can merge it

@reb-dev Done!

rokotyan:
@lee00678 Maybe I'm missing something but the changelog still shows me this
image

lee00678:
> @lee00678 Maybe I'm missing something but the changelog still shows me this image

Fixed.

reb-dev (approved):

PR #402, opened 2024-06-18
Tooltip: Sticky position and support for dynamic content (e.g. via React's `createPortal`)

rokotyan:
This PR will add support for sticky tooltip positioning:

https://github.com/f5/unovis/assets/755708/67e5b7f6-6746-4c7d-8a4b-d50a1c048b01

https://github.com/f5/unovis/assets/755708/b28e350d-315d-4754-9074-5b988f6e5261

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25633,7 +25633,6 @@
         "ng-packagr": "^12.0.0",
         "rollup": "^2.61.1",
         "rollup-plugin-typescript2": "^0.31.1",
-        "ttypescript": "^1.5.13",
         "webpack": "5.76.0",
         "zone.js": "^0.14.0"
       },
@@ -30872,7 +30871,6 @@
         "rollup": "^2.61.1",
         "rollup-plugin-typescript2": "^0.31.1",
         "tslib": "^2.3.1",
-        "ttypescript": "^1.5.13",
         "webpack": "5.76.0",
         "zone.js": "^0.14.0"
       },
--- a/packages/angular/package.json
+++ b/packages/angular/package.json
@@ -69,7 +69,6 @@
     "ng-packagr": "^12.0.0",
     "rollup": "^2.61.1",
     "rollup-plugin-typescript2": "^0.31.1",
-    "ttypescript": "^1.5.13",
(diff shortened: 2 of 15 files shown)


rokotyan:
@reb-dev  I think this is ready for review, but I'm having troubles finding time to update the docs to tell about these new features. If you think that either you or @lee00678 can do it, I would really appreciate it.

reb-dev:
@rokotyan Nice feature! I will add the docs within the next couple days.

reb-dev (changes requested):
@rokotyan When testing I noticed inconsistent behavior with crosshair when `followCursor` is `false`. The place where tooltip is hoverable seems to be random, and it sometimes follows the cursor even when crosshair isn't active.
 
It is difficult to showcase with a recording but here is an example of what I mean:


https://github.com/f5/unovis/assets/52078477/22130109-e22d-4dcf-a672-b64bda6f575d



I'm wondering if we can rework the logic so crosshair's tooltip can support this feature, or if we should automatically disable it?

rokotyan:
@reb-dev You're right, I haven't tested this use case, thanks for your attention to detail! I think Crosshair doesn't need a hoverable tooltip, so we can safely disable it. I'll update the PR.

reb-dev (dismissed):

reb-dev:
I made some changes to the Tooltip doc page and added the new info in the recent commit.

@lee00678 can you review? Approval needs to be from someone besides the last pusher.

lee00678:
@reb-dev looks good. Tested the dev examples and docs, both working as expected.

lee00678 (dismissed):

rokotyan:
Adding support for custom class names which is needed when using `createPortal` with multiple tooltips. Planning to finish this today.

rokotyan:
@reb-dev @lee00678 Pushed a few more commits. Ready for review again.

rokotyan:
New in the docs
image

image

image

reb-dev (approved):

rokotyan:
Rebased from `main`

lee00678 (approved):

PR #404, opened 2024-06-20
Graph: Supporting custom functions for node rendering; onRenderComplete callback

rokotyan:
This PR will allow the users to come up with their own node rendering functions, which will be a pretty common use case for everyone designing their own graphs.

It also add the `onRenderComplete` callback which allows you to do even more, e.g. add custom background / foreground graphics.

https://github.com/f5/unovis/assets/755708/cd36d1c5-891e-4af9-9c9e-c2a54dd5723a


packages/ts/src/components/graph/modules/node/label.ts
reb-dev, line 1: I don't think this file is being imported anywhere

lee00678:
Do we need to add documentation for this?

packages/ts/src/components/graph/index.ts
```diff
@@ -227,8 +234,22 @@ export class Graph<
...
+        this.config.onLayoutCalculated?.(datamodel.nodes, datamodel.links)
+      })
     }
 
+    // Redefine Zoom Behavior filter is specified in the config
+    // https://d3js.org/d3-zoom#zoom_filter
+    this._zoomBehavior.filter(
+      isFunction(zoomEventFilter)
+        ? zoomEventFilter
+        : (e: PointerEvent) => (!e.ctrlKey || e.type === 'wheel') && !e.button // Default filter
```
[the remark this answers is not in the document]
rokotyan: @reb-dev Good point, I haven't checked this part after I did a rebase

packages/ts/src/components/graph/modules/node/label.ts
rokotyan: Right, I thought I removed it but somehow it's still here. Thanks for pointing it out.

rokotyan:
@lee00678 Yes, definitely. And probably a new example. I feel like it's a pretty big feature so we might want to present it properly.

rokotyan:
Latest commits summary:

Reworking node selection and state update logic
- New Config property: `nodeSelectionHighlightMode` to control what to dim / highlight on node selection
- Fix: Node and link state was set after render
- Refactor: Removed unnecessary render calls
- Refactor: Renamed state-related rendering functions (updateNodesPartial, updateLinksPartial)

rokotyan:
@reb-dev (cc @lee00678) After some more local testing, I think I need to make the `updateNodesPartial` function configurable as well. Otherwise some of the styles might get overridden. Also to give you a heads up, I might need to move the logic around using `s.nodeSelectors.brushed` class there because it aligns better with what we have for grayed out and selected states.

rokotyan:
@reb-dev I've pushed a few new commits and tested them on my end, everything seems to be working well. I'm a bit concerned about [this change](https://github.com/f5/unovis/pull/404/commits/2170e17b0471b345583d39edb28f6a492bd24d5e) because I don't remember why I did it that way initially. It would be great if you could test it with your existing graphs and let me know if everything still looks good. You can use `1.5.0-nikita.1` version to test which has all the changes from this PR.

reb-dev (approved):
Sorry for the delay in merging, @rokotyan. Everything looks good!

rokotyan:
No worries, thanks @reb-dev

PR #419, opened 2024-07-24
Adding Visual Testing: Cypress + Percy

lee00678:
I'm setting the trigger for visual testing to happen when merge to main for now. I think we could also have Percy runs on a weekly basis just in case.

- [x] Installed Cypress and Percy for visual testing, built into our CI pipeline, triggers on merge to `main`
- [x] Added a random seed generator in our dev environment so the test can be consistent.


package.json
```diff
@@ -67,5 +67,8 @@
   },
   "lint-staged": {
     "*.{js,ts,jsx,tsx,svelte}": "eslint --quiet --cache --fix"
+  },
+  "dependencies": {
+    "seedrandom": "^3.0.5"
```
rokotyan, line 72: I think this shouldn't be a top-level dependency. Most likely you want it to be in `packages/dev`

packages/dev/cypress/support/commands.ts
```diff
@@ -0,0 +1,37 @@
+/// <reference types="cypress" />
```
rokotyan, line 1: Do we need this file at all?

packages/dev/src/utils/data.ts
```diff
@@ -1,5 +1,8 @@
 import { GenericDataRecord } from '@unovis/ts'
 import { sample } from './array'
+/* eslint-disable-next-line @typescript-eslint/no-var-requires, @typescript-eslint/naming-convention */
+const SeedRandom = require('seedrandom')
```
rokotyan, line 4: Is there a problem with using the `import` syntax?

packages/dev/src/utils/data.ts
```diff
@@ -1,5 +1,8 @@
...
+export const rng = new SeedRandom('hello.')
```
rokotyan, line 5: Maybe: `hello.` → `unovis` 🙂

rokotyan: Also what `rng` stands for? I would try to pick a more explanatory name to make the code easier to read

packages/dev/cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
```
rokotyan, line 1: Do we need this file at all?

packages/dev/cypress/e2e/unovis.cy.ts
```diff
@@ -0,0 +1,358 @@
...
+  })
+
+  it('Load homepage', () => {
+    cy.percySnapshot('Homepage test')
+  })
+
+  it('Basic Annotation', () => {
+    cy.contains('Basic Annotations').click()
+    cy.wait(3000)
+    cy.percySnapshot('Basic Annotation', { scope: '.exampleViewer' })
```
rokotyan, line 13: I would move `{ scope: '.exampleViewer' }` to a dedicated variables since it is repeated so many times

packages/dev/cypress/e2e/unovis.cy.ts
```diff
@@ -0,0 +1,358 @@
...
+    cy.visit('/')
+  })
+
+  it('Load homepage', () => {
+    cy.percySnapshot('Homepage test')
+  })
+
+  it('Basic Annotation', () => {
+    cy.contains('Basic Annotations').click()
+    cy.wait(3000)
```
rokotyan, line 12: Is there a specific reason to wait 3000 ms?

packages/dev/cypress.config.ts
```diff
@@ -0,0 +1,20 @@
+import { defineConfig } from 'cypress'
+// import webpackConfig from './webpack.config'
```
rokotyan, line 2: Unnecessary comment

packages/dev/cypress/e2e/unovis.cy.ts
```diff
@@ -0,0 +1,358 @@
...
+    cy.percySnapshot('Scale by Domain', { scope: '.exampleViewer' })
+  })
+
+  it('Stacked Bar with Labels', () => {
+    cy.contains('Stacked Bar with Labels').click()
+    cy.wait(3000)
+    cy.percySnapshot('Stacked Bar with Labels', { scope: '.exampleViewer' })
+  })
+})
+
```
rokotyan, line 358: We can try generating this file automatically based on the example files. Technically they have the required information, like the description text.

rokotyan (commented):
This is great! Can you please also provide some steps how to set it up locally?

lee00678:
> This is great! Can you please also provide some steps how to set it up locally?

Running this locally is a little tricky. You'd have to have a Percy account and a token to be able to do that. See [this](https://www.browserstack.com/docs/percy/overview/basics). It is possible to have a free personal account, which is what I did when I was testing this out. You need to create a project and it will generate a PERCY_TOKEN. You will then set it as part of your local env.

packages/dev/src/utils/data.ts
lee00678: > Also what `rng` stands for? I would try to pick a more explanatory name to make the code easier to read

Random number generator. I thought it's pretty standard. I could update it to it's full name.

packages/dev/cypress/e2e/unovis.cy.ts
lee00678: > Is there a specific reason to wait 3000 ms?

Some page has animations. This way we wait till everything is fully rendered before taking the screenshot.

packages/dev/cypress/e2e/unovis.cy.ts
lee00678: > We can try generating this file automatically based on the example files. Technically they have the required information, like the description text.

I suppose we could, but some page require additional settings. For example the `API Endpoints Tree` requires a bigger minHeight, we'd have to manually set things like that for specific test cases. Do you have a workaround for cases like this?

package.json
lee00678: Good call.

reb-dev (approved):

packages/dev/cypress/support/commands.ts
lee00678: It turns out we do need this file. Without it Cypress can't resolve `/commands`. See [here](https://github.com/f5/unovis/actions/runs/10203526794/job/28230045052)

packages/dev/cypress/e2e/unovis.cy.ts
rokotyan: Should we add `duration` as a query string argument to the pages? I think this way you can speed up the testing significantly.

packages/dev/cypress/e2e/unovis.cy.ts
rokotyan: You can have a map of additional settings in the testing file to apply on the go if needed. I can explain more if you want, just let me know.

lee00678: Do you mean to have an array or load a separate file which contains 
```ts
testingArray = [
  {name: 'Single Container', duration:3000, additionalParams: {width: [1300]}},
  {name: 'Sankey', duration:3000},  ...
]
```
And iterate through the array inside the `unovis.cy.ts`?

packages/dev/cypress/e2e/unovis.cy.ts
lee00678: > Should we add `duration` as a query string argument to the pages? I think this way you can speed up the testing significantly.

What do you mean? Add something like `?duration=3000` to the url?

rokotyan: More like `?duration=0` to render charts without animation

packages/dev/cypress/e2e/unovis.cy.ts
rokotyan: Take a look at `unovis/packages/dev/src/examples/index.ts`, I was thinking about something similar to generate this test file automatically. You'll need to "compile" it using Webpack but I think it will be a reasonable trade-off.

So the script will:
- Batch import all the `index.tsx` files in the examples folder;
- You'll be able to get the required information from the imports (titles, ...);
- You can define additional testing configuration in the `unovis.cy.ts` file and match the corresponding examples by name or path. Or you can define it in the `index.tsx` files of the examples themsevles.

packages/dev/cypress/e2e/unovis.cy.ts
lee00678: > More like `?duration=0` to render charts without animation

That's a good idea.

packages/dev/cypress/e2e/unovis.cy.ts
lee00678: @rokotyan I will give it a try. Is this PR getting too big? Should we have a separate PR for this and the `duration` params?

packages/dev/src/examples/auxiliary/annotations/basic-annotations/index.tsx
lee00678, line 1: @rokotyan @reb-dev Updating this as a proof of concept to see this is the right approach. I will update the rest of the files if this looks good.

packages/dev/cypress/e2e/unovis.cy.ts
lee00678: > More like `?duration=0` to render charts without animation

See a POC [here](https://github.com/f5/unovis/pull/419/files#r1706006409)

packages/dev/cypress/e2e/unovis.cy.ts
rokotyan: Up to you. I would do it here, but I don't see any problem with merging this PR first and having another one

packages/dev/src/examples/auxiliary/annotations/basic-annotations/index.tsx
rokotyan: Nice! I think I would try doing it in `packages/dev/src/components/ExampleViewer/index.tsx` and passing as a prop, or defining a dedicated React hook that can be imported to avoid code duplication.

packages/dev/src/components/ExampleViewer/index.tsx
lee00678, line 1: @rokotyan @reb-dev updated by passing `duration` as props.

lee00678: @reb-dev the latest push should fully implement the `duration` props for all dev files. I put the urls in a separate file, we can fully implement the autogen for testing urls in a separate PR.

reb-dev (approved):

sumitkumar25:
@lee00678 dont enable percy run on merge to master. it will eat up the screenshot quota. 
1. i would recommend a weekly run.
2. We should check if we can have a percy run before a new release version is published.

lee00678:
> @lee00678 dont enable percy run on merge to master. it will eat up the screenshot quota.
> 
> 1. i would recommend a weekly run.
> 2. We should check if we can have a percy run before a new release version is published.

It has a weekly run. Removed the trigger for merge to main.

reb-dev (approved):

Issue #442, opened 2024-09-05
Incidental Changes around Licence.txt


curran:
It seems that a bunch of files automatically get modified when you run `npm install`. This causes friction for development, since these changes either must be manually unstaged from Git for a clean commit, or must be kept up to date. I'm wondering what is the best approach for dealing with this? For now I'll do the following

 * Start with a PR that just updates the license.txt files and package-lock
 * Stack a PR on top of that with my new changes

curran:
Aha! A workflow that can work:

 * `git checkout main`
 * `git pull`
 * `npm install`
 * `git checkout .` <-- unstage the changes from auto-generated files
 * Work as usual

PR #443, opened 2024-09-05
Misc | Update generated files: Run npm install to update generated files

curran:
Related to https://github.com/f5/unovis/issues/442


curran:

PR #457, opened 2024-10-08
Enable sourcemaps in dev

### LINKED ISSUE #456: Support Sourcemaps for Dev
As a newcomer to the Unovis codebase, I want to be able to know the file and line number from which an error originated, so that I can figure out the cause.

Current behavior: in `npm run dev`, there seems to be no sourcemaps configured, so when an error occurs, it resolves to `bundle.js`, which is useless. Example:

image

Desired behavior: in `npm run dev`, the environment leverages sourcemaps to tell you where the error is coming from.

curran:
Closes #456 

Before:

image

After:

image

### DIFF
--- a/packages/dev/webpack.config.js
+++ b/packages/dev/webpack.config.js
@@ -8,7 +8,7 @@ const path = require('path')
 const isDevelopment = process.env.NODE_ENV !== 'production'
 module.exports = {
   entry: './src/index.tsx',
-  devtool: 'source-map',
+  devtool: isDevelopment ? 'eval-source-map' : 'source-map',
   mode: isDevelopment ? 'development' : 'production',
   output: {
     path: path.resolve(__dirname, 'dist'),


packages/dev/webpack.config.js
```diff
@@ -8,7 +8,7 @@ const path = require('path')
 const isDevelopment = process.env.NODE_ENV !== 'production'
 module.exports = {
   entry: './src/index.tsx',
-  devtool: 'source-map',
+  devtool: isDevelopment ? 'eval-source-map' : 'source-map',
```
rokotyan, line 11: 🤔 I remember having no issues with debugging Unovis from the dev app. Was the error in the console also leading to the obfuscated bundle? 

Also the `source-map` option generally produces better results that 'eval-source-map' and is more computationally expensive to calculate. Do you know why the `eval-source-map` works here better here?

curran: This change changes the error message from pointing to `bundle.js` to pointing to something from `node_modules`. That's arguably an improvement, but actually it's not that much better, since we still are not getting the line numbers for the original error. Happy not to merge this, maybe it's just a weird case I'm dealing with. Thanks for reviewing!

curran: But overall, the source maps don't seem to be working.

For example, when I add this change:

image

I get this error:

image

The source maps don't seem to be doing their job. Instead of `bundle.js` it should resolve to the specific line number and source file, but it does not.

curran: After changing it to `eval-source-map`, the line numbers DO resolve correctly.

image

lee00678:
Hi @rokotyan @curran, what's the status of this? Do we want this merged in?

Issue #455, opened 2024-10-08
Resize Observer Error on Vertical Resize


curran:
While working with Unovis components, I notice that this error occurs frequently.

<img width="1102" alt="image" src="https://github.com/user-attachments/assets/7e9cb401-c5cb-496c-a801-6070302f02b3">

Reproduction:

 * Change `<VisSingleContainer height={400}>` to `<VisSingleContainer>` on line 11 of `packages/dev/src/examples/misc/donut/donut-empty-segments/index.tsx`
 * Open that example at http://localhost:9500/examples/Donut/Donut:%20Empty%20Segments
 * Resize vertically
 * See the error happen

The docs say:

>By default, Container automatically fits to the size of the parent element.

so, I would expect this to work without errors.

curran:
After enabling sourcemaps in #457 , I see this:

<img width="1494" alt="image" src="https://github.com/user-attachments/assets/761ab395-8fde-45b0-912e-05b3d9bea29c">

Related issue:

https://github.com/webpack/webpack-dev-server/issues/4812

It seems that what's happening is that the Webpack dev server is intercepting the error, which is coming from within the Unovis codebase

curran:
Actually, this error is happening on the current `main`, even without changing the height of the donut chart.

curran:
Researching this, the most common way to solve it is to throttle or debounce the updates. That's something I can try, which seems like pretty low hanging fruit!

curran:
Oh wow there's already a dependency here for https://github.com/niksy/throttle-debounce

curran:
Some things I tried that didn't work:

 * Debouncing the rendering
 * Setting a flag to ignore resize observer events when rendering is taking place

The error still happened after either of those.

I didn't open any PRs with these changes since they didn't work. I'd like to get to the bottom of this one, since it's the starting point for local testing of vertical resize.

I wonder if there's some workaround where I could avoid solving this but also enable vertical resizing of the dev page for

 * #454

PR #458, opened 2024-10-08
WIP fix resize observer bug

curran:
WIP towards #455

### DIFF
--- a/packages/ts/src/core/container/index.ts
+++ b/packages/ts/src/core/container/index.ts
@@ -137,6 +137,10 @@ export class ContainerCore {
 
   protected _setUpResizeObserver (): void {
     if (this._resizeObserver) return
+
+    // eslint-disable-next-line
+    console.log('Setting up Resize Observer on element', this._container)
+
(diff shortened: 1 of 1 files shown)


curran:
Current status: uncovered the fact that multiple resize observers are set up on the same element.

image

It looks like the constructor is invoked twice for the exact same element.

rokotyan:
> It looks like the constructor is invoked twice for the exact same element.

This is probably due to React rendering components twice in the dev [StrictMode](https://react.dev/reference/react/StrictMode#strictmode), which is a known behavior and usually not an issue — Unovis container components should disconnect the observer when destroyed.

rokotyan:
I was able to reproduce this problem after setting "Show scroll bars" to "Always" in the macOS settings.

image

rokotyan:
#463 should fix this

50rayn:
Is this PR still actual ?

rokotyan:
@lee00678 I think we can close this one. I've fixed this issue in another PR.

PR #463, opened 2024-10-11
Core | Container: Improve ResizeObserver handling

### LINKED ISSUE #455: Resize Observer Error on Vertical Resize

rokotyan:
This PR fixes #455, a problem with ResizeObserver throwing an error when scrollbars appear/disappear. Solving this by delaying the resize by one frame.

The only tradeoff is scrollbar flickering when changing the size, but it only appears when "Show scroll bars" setting is set to "Always" on macOS (not sure about the other platforms).

image

https://github.com/user-attachments/assets/b6f15fe6-acad-484b-9712-bc51e48ca864

Changelog:
- Use requestAnimationFrame to avoid multiple resize events when scrollbars appear/disappear
- Cancel previous animation frame IDs before scheduling new ones
- Renamed `_requestedAnimationFrame` to `_renderAnimationFrameId`

### DIFF
--- a/packages/dev/src/examples/misc/donut/donut-full-height/index.tsx
+++ b/packages/dev/src/examples/misc/donut/donut-full-height/index.tsx
@@ -1,0 +1,22 @@
+import React from 'react'
+import { VisSingleContainer, VisDonut } from '@unovis/react'
+import { ExampleViewerDurationProps } from '@src/components/ExampleViewer/index'
+
+export const title = 'Donut: Full Height'
+export const subTitle = 'Testing the resize behavior'
+
+export const component = (props: ExampleViewerDurationProps): JSX.Element => {
+  const data = [3, 2, 5, 4, 0, 1]
+  return (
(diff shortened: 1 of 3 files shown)


rokotyan:
@lee00678 I've pushed a new example for this

lee00678:
> @lee00678 I've pushed a new example for this

Thanks for the clarification.

curran:
Hooray!

PR #422, opened 2024-07-29
feat: Solid support 

hngngn:
This PR adds SolidJS support to Unovis, expanding framework compatibility.

PR #469, opened 2024-10-21
Solid Integration

rokotyan:
A clone of https://github.com/f5/unovis/pull/422, but rebased from `main` instead of merging. Let's see if the build passes.


rokotyan:
There is one more problem with the build I've found. All Unovis libraries preserve the source directory structure in the dist, making it possible to import components individually, like `import { VisAnnotations } from '@unovis/react/components/area'`:
image

The Solid build doesn't have that, and it seems to be bundling all of `@unovis/ts` plus its dependencies. This will have a negative effect on the final bundle size of the apps that use `@unovis/solid`.
image

I think you can take a look at the `vite.config.ts` from `packages/vue` and use it as an example of the build configuration.

lee00678:
> @hngngn @lee00678
> 
> I was able to make the solid build work, but the website build is failing because it can't find the examples for some reason. If you can take a look into this, it would be great. image

Yea, this happened to me during my test a couple of days ago as well. (https://github.com/f5/unovis/pull/475#issuecomment-2435549646) Seems to be this is caused by solid, so we may need some insights from @hngngn

hngngn:
Sorry for the late reply—I've been pretty busy lately. I'll look into the issue and report back ASAP.

hngngn:
I've already pushed a fix for both problems you're facing. Can you test it? @rokotyan @lee00678

hngngn:
I've pushed a fix. Could you test it?

lee00678:
@rokotyan do you mind pushing the latest fix so I can do a last round of testing before merging?

rokotyan:
@lee00678 Done. Amended the main commit and did a force push.

lee00678:
> I think you can take a look at the `vite.config.ts` from `packages/vue` and use it as an example of the build configuration.

This looks good!

lee00678:
@hngngn Thanks for this contribution! This will be released in 1.5.0.

lee00678 (approved):

PR #476, opened 2024-10-25
Component | Tooltip | Crosshair: Add default horizontal placement for Crosshair

rokotyan:
Changelog:
- Add `_defaultHorizontalPlacement` property to store the default horizontal placement for Tooltip
- Use `_getHorizontalPlacement()` to get the configured horizontal placement, falling back to the default if not set
- Set the default horizontal placement to `Position.Right` when used with Crosshair


packages/ts/src/components/crosshair/index.ts
```diff
@@ -179,6 +182,7 @@ export class Crosshair<Datum> extends XYComponentCore<Datum, CrosshairConfigInte
     const content = config.template(this.datum, this.xScale.invert(this.x))
     // Force set `followCursor` to `true` because we don't want Crosshair's tooltip to be hoverable
     tooltip.config.followCursor = true
+    tooltip.config.horizontalPlacement = tooltip.config.horizontalPlacement || Position.Right
```
lee00678, line 185: I think the original behavior will calculate the position based on the x position: https://github.com/f5/unovis/pull/402/files#diff-349d9b2f94ecdc1fe4615a4979bf0fee791c4006b0fed078fd416c31f90f0eb1L118

rokotyan: @lee00678 Good point, I'll update the PR

rokotyan: @lee00678 Fixed this problem but forgot about about one other thing. Will keep you posted

rokotyan:
@lee00678 This is ready for review

https://github.com/user-attachments/assets/efc5756f-1304-4343-b4b9-c03327474ff4

lee00678 (approved):

Issue #484, opened 2024-11-19
Svelte 5 Infinite Effect / Reactive Loop


pingu-codes:
Using svelte 5 with unovis causes an infinite reactive loop if more than 1 x/y component is rendered. The cause of this is the `component` action set within the xy container context:

```ts
setContext('component', () => ({
		update: (c: XYComponentCore<Datum>) => {
			config.components = [...config.components, c];
		},
		destroy: () => {
			config.components = config.components.filter((c) => !c.isDestroyed());
		}
	}));
```

Due to svelte 5's deeper reactivity the update function is called when data is added to one of the child line components, as a result of this the config.components pushes a duplicate component to `config.components` and an infinite loop begins and pushes duplicates to config.components.

A fix for this is simply preventing the update function from writing duplicates to `config.components`
```ts
setContext('component', () => ({
		update: (c: XYComponentCore<Datum>) => {
			if (config.components?.some((e) => e === c)) {
				return;
			}

			config.components = [...config.components, c];
		},
		destroy: () => {
			config.components = config.components.filter((c) => !c.isDestroyed());
		}
	}));
```

rokotyan:
@pingu-codes Thanks for figuring out the problem; we really appreciate it! We'll try it out soon and release a new version.

lee00678:
@pingu-codes Thanks for the suggestion. I just released `1.5.0-beta.1` with your suggested fix. The browser renders a lot faster now, but I'm still seeing the infinite loop issue when using `axis` and `annotations` (I tried it with most of our existing components, others seem to be working. Here's the branch I used to publish `1.5.0-beta.1`: [qian/svelte-bug-fix](https://github.com/f5/unovis/tree/qian/svelte-bug-fix). Any suggestions?

pingu-codes:
Hey @lee00678, it looks like it's largely the same problem for all of them, I suspect it's the same for all the components with separate `lifetime` actions. I'm not very familiar with the inner workings of `@unovis/ts` but for the axis specifically something within `chart?.updateContainer({ ...config, ...($$restProps as XYContainerConfigInterface<Datum>) });` is causing the update action to be recalled. From my testing adding code like the following should fix the reactivity loop

```ts
if (config[`${e.__type__}Axis`] === c)
    return;
}
// ...

if (config.crosshair === c) {
    return;
}
```

pingu-codes:
[...]

PR #487, opened 2024-11-22
Svelte: Fix reactivity loops & crashes

pingu-codes:
This PR implements the changes discussed in #484 which fixes bugs that occur from Svelte 5's new reactivity system.

### DIFF
--- a/packages/svelte/src/containers/xy-container/xy-container.svelte
+++ b/packages/svelte/src/containers/xy-container/xy-container.svelte
@@ -38,6 +38,8 @@
   let ref: HTMLDivElement
 
   $: chart?.setData(data, true)
+  // eslint-disable-next-line no-self-assign
+  $: config.components = config.components
   $: chart?.updateContainer({ ...config, ...($$restProps as XYContainerConfigInterface<Datum>) })
 
   onMount(() => {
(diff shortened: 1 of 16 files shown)


packages/svelte/src/containers/xy-container/xy-container.svelte
```diff
@@ -46,26 +48,39 @@
   })
 
   setContext('component', () => ({
-    update: (c: XYComponentCore<Datum>) => { config.components = [...config.components, c] },
+    update: (c: XYComponentCore<Datum>) => {
+      if (config.components?.some((e) => e === c)) return
```
rokotyan, line 52: We can probably use the `includes()` function here instead

packages/svelte/src/containers/xy-container/xy-container.svelte
```diff
@@ -38,6 +38,8 @@
   let ref: HTMLDivElement
 
   $: chart?.setData(data, true)
+  // eslint-disable-next-line no-self-assign
+  $: config.components = config.components
```
rokotyan, line 42: Do you know if there are there any drawbacks of doing this, e.g. extra re-renders when nothing changes or something like that?

rokotyan (commented):
Thanks for the contribution! Let a few minor comments

packages/svelte/src/containers/xy-container/xy-container.svelte
pingu-codes: I'm not sure it actually fixes what it was originally supposed to. My original issue was that the chart wasn't rerendering when child component data changes but this doesn't actually fix that. I'm not sure if it's intended that when a child components data is set the whole graph rerenders.

If it is it'll need a different fix and I can look into it lmk :)

packages/svelte/src/containers/xy-container/xy-container.svelte
pingu-codes: fixed

pingu-codes:

lee00678:
Just released [1.5.0-svelte.0](https://www.npmjs.com/package/@unovis/ts/v/1.5.0-svelte.0). I tested it with a simple Svelte 5 project, seems to be working for me.

lee00678:
@pingu-codes This PR looks good. Do you mind cleaning up the commit history (part of my fault as well)? You can either squash it or overwrite the history and do a force push to the branch.

pingu-codes:
Yep, just squashed the commits!

pingu-codes:
Do you think I should remove the changes to the package-lock.json on website too as it seems to include a lot of changes not really related to the fix?

50rayn:
I would rebase onto the latest `main` branch. After that, discard changes from `package-lock.json` and rerun an `npm ci`. And recheck if everything is ok and works as expected. Why `npm ci`, because `npm i` adds new dependencies and updates dependencies on a project.

lee00678:
> Do you think I should remove the changes to the package-lock.json on website too as it seems to include a lot of changes not really related to the fix?

Yea, I would rebase from `main`. We did a major website upgrade so there will be a lot of updates in the website `package-lock.json`

lee00678 (approved):

pingu-codes:
Did you handle the changes with your push @lee00678 sorry I've been really busy the last few days. I'm pretty bad at github tbh (mainly use graphite at my work) so I don't want to break anything if you've made the changes you need 😅

sumitkumar25 (approved):

PR #489, opened 2024-11-25
Misc: Integrate the bumpp

50rayn:
Just saw the bash script `update-version.sh`, which help to synchronize the same version across all the _package.json_ files in all packages, and recall the [bumpp](https://github.com/antfu-collective/bumpp) module. This module do the same and is high configurable, which allows to configure the commit message, tag name or even push after rising the version or not. All configuration can be done from CLI or the `bump.config.ts`. I created the configuration file as similar as the `update-version.sh`. Also observed that for each _Release: X.X.X_ commit is just raised the version and commit with the same commit message and tag template, these details are also added in config file.

2024-11-26 at 12 52 31


rokotyan:
@50rayn This is a great tool! But I'm having a problem with it. Every time I'm trying to update the version, it get's stuck in an endless loop saying  `Failed to locate the previous tag`. Do you know how to fix it? I've tried creating a git tag manually, but it didn't help.

https://github.com/user-attachments/assets/334ac239-4af7-4098-9512-5582d951dacd

50rayn:
@rokotyan Played around and and found `--ignore-scripts` option based on [docs](https://github.com/JS-DevTools/version-bump-prompt?tab=readme-ov-file#version-scripts). So it was fixed. But I'd rename the script from package.json from `version` to `bumpp` or smth else.

rokotyan:
> I tried locally to rename the version to bumpp, and it worked without setting the ignoreScripts flag.
Tried this again, finally. Renamed `verison` to `bump` in `package.json` and  worked like a charm. @lee00678 I think it's the right thing to do.

50rayn:
@rokotyan If the script was renamed to bump, the option `ignoreScripts` from the config can be removed, since it's false by default.

rokotyan:
@50rayn Yep, I think we want to keep the scripts. I like the automatic tagging feature

lee00678 (approved):
@50rayn thanks for the contribution. Sorry it took a while.

lee00678:
Hi @50rayn, we just used this feature to release the latest Unovis. Very nice feature. However, I noticed when run `npm run version`, the peerDependency `@unovis/ts` doesn't get updated to the latest version. I think we may need a postbump script for that?

50rayn:
Hello @lee00678. I'm not sure that `bumpp` doesn't have such functionality of upgrading the version from _peerDependency_.

PR #493, opened 2024-11-27
Core: Fix the nullish coalescing operator in `_updateScalesRange`

50rayn:
I got this interesting warning when playing around with the project build.

```
src/containers/xy-container/index.ts: The "??" operator here will always return the left operand
```

And the compiler was right. In case `config.padding.bottom` will be `undefined`, subtraction will get NaN. The operator will be applied to the result of the subtraction but not the `config.padding.bottom`. So the fix was easy, just wrapping up with paratheses.


rokotyan:
@50rayn You're right! I wonder how did we miss it. Thanks

rokotyan:
@50rayn Cool, it's something I wanted to do for a long time

lee00678 (approved):

Issue #440, opened 2024-09-03
Regression in custom SVG defs for Containers in Versions > 1.4.1


flschweiger:
It seems that using custom SVG defs with containers stopped functioning properly in versions later than 1.4.1. For instance, applying custom SVGs as colors for a `StackedBar` within an `XY` container no longer renders as expected.

Upon comparing versions 1.4.1 and 1.4.2, I noticed changes in the implementation within [packages/ts/src/core/container/index.ts](https://github.com/f5/unovis/compare/1.4.1...1.4.2#diff-1efdbf519e7440940c9ed9ec1051a0e177fe1c8e28ba08780cbbe224071d40fb). A quick review suggests that these custom SVGs might not be added to the `svg` element anymore.

I'll try to investigate this issue further when I get a chance, but I wanted to bring it to your attention in the meantime 🙂

Thanks for maintaining such a great library!

rokotyan:
@flschweiger Thank you for reporting the issue! Can you please show us how you have set up your chart? I have not been able to reproduce the problem with my configuration.

stevenhurth:
I've found that if I specify the svg def in the template html (vue 3), it will work. However, using the svgDefs prop does not work.

rokotyan:
@stevenhurth Can you please confirm that you're using the latest `@unovis/ts` and `@unovis/vue` versions (1.4.4)?

MikeBellika:
> @stevenhurth Can you please confirm that you're using the latest `@unovis/ts` and `@unovis/vue` versions (1.4.4)?

@rokotyan I can confirm it's still an issue with 1.4.4 (and 1.5 beta). Here's a repro:
1.4.4: https://stackblitz.com/edit/github-vso9jp?file=package.json
![image](https://github.com/user-attachments/assets/92d292db-554c-4a4c-a897-6b30efea49c7)


1.4.0: https://stackblitz.com/edit/github-vso9jp-1itphx?file=package.json
![image](https://github.com/user-attachments/assets/d4f5ffbe-718b-4bb5-bca7-b36c9839ae87)

rokotyan:
Thanks @MikeBellika! The fix will be officially released soon, in the meantime you can use `1.4.5-beta.1`

rokotyan:
Version `1.4.5` has been released with the fix for this issue

PR #494, opened 2024-11-27
Core | Container: Moving svgDefs update to `updateContainer`

rokotyan:
Fixes https://github.com/f5/unovis/issues/440

Moving svgDefs update to `updateContainer` from `_render`, because `prevConfig` can be reassigned multiple times before render happens leading to `config` and `prevConfig` to be the same


lee00678:
@rokotyan I can merge this right?

rokotyan:
@lee00678 Yes, we've already released 1.4.5 with it

lee00678 (approved):

PR #496, opened 2024-11-29
Misc: Add `.nvmrc` file and set packageManager

50rayn:
Add the `.nvmrc` configuration file to ensure that the correct node version is used.
I followed the configuration from `pull_request.yml`, there is also a set 20th.
I also have other projects that are set to other versions, and I'm using [fnm](https://github.com/Schniz/fnm), when I'm switching the projects it switches automatically to the necessary one. This will ensure that I use the expected Node version and install the "correct" modules.

Besides the nvm configuration, I set to all package.json files the packageManager. As with the nvm, it will ensure that the modules are installed with the corresponding _packageManager_ and version.


rokotyan:
Cool, thanks! We will give it a try and merge if everything works well

rokotyan (approved):
Worked well on my end

@50rayn In your opinion, what are the pros and cons of specifying the exact `packageManager` version, i.e. `npm@10.9.1`, instead of just `npm@10`?

50rayn:
We can't specify the `npm@10` only. It is necessary to specify the exact version since there is a required pattern `(npm|pnpm|yarn)@\d+\.\d+\.\d+(-.+)?`. So, based on the [documentation](https://nodejs.org/api/packages.html#packagemanager), it will ensure the same dev environment across the team. By the useful toll ([@anftu/ni](https://github.com/antfu-collective/ni)). In my opinion and practice:

**Cons:**
All dev will need to configure the same npm version (or pnpm soon :D). But I don't see this as a big problem since there are a lot of tools, like [nvm](https://github.com/nvm-sh/nvm), [fnm](https://github.com/Schniz/fnm), [corepack](https://nodejs.org/api/corepack.html), which will help dev to switch between required packageManager easily.

**Pros:**
Specifying the major version is not enough, I would say at least a minor version is needed. Some modules are sensible to minor versions, or we can use some modern flags from npm, which appeared later in npm but are not available from the beginning of the major version.

rokotyan:
@50rayn Thanks for the explanation. I think it'll be a good addition to our dev set up.

lee00678:
Hi @50rayn, thanks for your contribution. Now that we've released 1.5, we are getting ready to merge this PR. Do you mind rebase main and solve the merge conflicts? Thanks.

50rayn:
No problem @lee00678 . Will be soon 🚀

lee00678 (approved):

PR #497, opened 2024-12-03
Vue: update `useForwardProps` function to latest implementation

sadeghbarati:
cc: @zernonia 🙌

### DIFF
--- a/packages/vue/src/utils/props.ts
+++ b/packages/vue/src/utils/props.ts
@@ -1,5 +1,11 @@
 import { isEqual } from '@unovis/ts'
-import { ComponentInternalInstance, camelize, computed, getCurrentInstance } from 'vue'
+import { type MaybeRefOrGetter, camelize, computed, getCurrentInstance, toRef } from 'vue'
+
+interface PropOptions {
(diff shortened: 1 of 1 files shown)


sadeghbarati:

rokotyan:
@sadeghbarati Thanks for the contribution! Could you please tell what the goal of these changes is?

sadeghbarati:
Hi @rokotyan, sorry I didn't explain fully. I apologize 🙌

`useForwardProps` utility is ported from `radix-vue`, and that function is updated, so I thought it would be better to include it here, too

https://github.com/unovue/radix-vue/blob/main/packages/radix-vue/src/shared/useForwardProps.ts

I hope zernonia could clarify whether this change is needed or not

zernonia:
Thanks for the PR @sadeghbarati , however in this particular instance I adapted `useForwardProps` specifically for `unovis`. 
Some component having trouble resolving the complex type, and required to use attribute to pass on the specific props. Thus you can see `const attrs = vm.attrs`.  

I don't think we need to make this changes 😁

PR #502, opened 2024-12-06
Half Donut

curran:
This PR introduces support for a "half donut" chart.

Related:

 * https://github.com/ExaForce/unovis/pull/3

image

image

image

image

Summary of changes:

 * Export constants that define half donut angle ranges
 * Add a new example that cycles through orientations (top, right, bottom, left) and also lets you resize dynamically  (height > width, height < width)
 * Make the scaling work for all 8 cases (top, right, bottom, left) X (height > width, height < width)

Based on a previous attempt in

 * https://github.com/f5/unovis/pull/444/files


curran:
Closing as this one is dated

Issue #347, opened 2024-03-11
No support for angular@17


juanmluces:
It appears that there is no support for angular v17, in the peer dependancies it is specified that the allowed verions are between v12 and v16 

``` "@angular/common": "12 - 16",
    "@angular/compiler": "12 - 16",
    "@angular/core": "12 - 16" 
```

![image](https://github.com/f5/unovis/assets/63096899/a78d3ae4-2376-42ed-bbba-cfec14659a8b)

PR #505, opened 2024-12-09
React 19 and Angular 19 support

rokotyan:
Adding React 19 and Angular 19 to the list of supported versions. 
- A quick dev app test shows that React 19 is working fine. 
- Angular historically has a good backwards compatibility, so I also assume it'll just work. I've checked out the release notes and didn't find any deprecations there.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28630,9 +28630,9 @@
         "zone.js": "^0.14.0"
       },
       "peerDependencies": {
-        "@angular/common": "12 - 18",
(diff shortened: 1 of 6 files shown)


50rayn:
This PR will not also close the issue #347?

rokotyan:
> This PR will not also close the issue #347?

This has been fixed a while ago, thanks for bringing it up. @lee00678 We should close that issue.

lee00678 (dismissed):

lee00678 (approved):

PR #509, opened 2024-12-13
Component | Donut: Half Donut

curran:
Adds support for half donut including:

 * Scaling to fit the geometry in the container
 * Applying label offsets automatically if `centralLabelsOffsetY` and `centralLabelsOffsetX` are not defined

image

image

image

image


lee00678 (approved):

curran:
Thanks!

PR #540, opened 2025-03-04
Dev | Fix: TypeScript import aliases

### LINKED ISSUE #539: React Package Types Unavailable
Image

I'm noticing that in VSCode (open to `unovis` as the root of the project), I'm not getting correct type hints for the `VisTreemap` export from `@unovis/react`. This adds friction during development. Ideally the types / linting would be correct.

curran:
Closes #539

### DIFF
--- a/packages/dev/tsconfig.json
+++ b/packages/dev/tsconfig.json
@@ -10,6 +10,8 @@
     "skipLibCheck": false,
     "paths": {
       "@src/*": ["packages/dev/src/*"],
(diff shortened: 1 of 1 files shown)


lee00678 (approved):

rokotyan:
@lee00678 After merging this I'm getting errors thorough the whole dev app: 

image
image



I think it's related to React version mismatch but we want to keep the older react version in `packages/react`. So I suggest to revert this change.

PR #542, opened 2025-03-20
fix: make fontSize optional

TheAlexLichter:
[According to the docs](https://unovis.dev/docs/auxiliary/Annotations#styled-text), `fontSize` is an optional property. This should be reflected in the types as well.

### DIFF
--- a/packages/ts/src/components/annotations/types.ts
+++ b/packages/ts/src/components/annotations/types.ts
@@ -1,9 +1,9 @@
-import { LengthUnit } from 'types/misc'
+import { LengthUnit, WithOptional } from 'types/misc'
 import { UnovisText, UnovisTextOptions } from 'types/text'
 
 
 export type AnnotationItem = Omit<UnovisTextOptions, 'x'|'y'|'width'> & {
-  content: string | UnovisText | UnovisText[];
+  content: string | WithOptional<UnovisText, 'fontSize'> | WithOptional<UnovisText, 'fontSize'>[];
   subject?: AnnotationSubject;
   x?: LengthUnit;
   y?: LengthUnit;
--- a/packages/ts/src/types/text.ts
+++ b/packages/ts/src/types/text.ts
@@ -24,8 +24,8 @@ export enum TextAlign {
 export type UnovisText = {
   // The text content to be displayed.
   text: string;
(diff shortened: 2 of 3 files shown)


TheAlexLichter:

rokotyan:
@TheAlexLichter (@lee00678)

I'm not confident that this property is optional, how did you come to your conclusion?

Looking at the `getWrappedText` function, the `fontSize` property doesn't seem to be optional 
image

lee00678:
> @TheAlexLichter (@lee00678)
> 
> I'm not confident that this property is optional, how did you come to your conclusion?
> 
> Looking at the `getWrappedText` function, the `fontSize` property doesn't seem to be optional image

I haven't got a chance to review this PR. But I'm also open to make it an optional property.

TheAlexLichter:
@rokotyan Given this example in the docs (and others) not adding explicit `fontSize` values.

https://github.com/f5/unovis/blob/79f45ddcec120af181f301bcc8d6af80f913a5a1/packages/website/docs/auxiliary/Annotations.mdx?plain=1#L102-L116

rokotyan:
@TheAlexLichter I see. It's optional only for the Annotations component, so the type should be this:

image

rokotyan:
Thanks for bringing up this inconsistency!

TheAlexLichter:
@rokotyan Good point! Updated my PR to reflect that! ☺️

rokotyan:
Thanks!

lee00678:
@TheAlexLichter could you please take a look at this: https://github.com/f5/unovis/actions/runs/14013488915/job/39393838730?pr=542? Thanks

rokotyan:
I've checked this again, and found that merging with default happens [here](https://github.com/TheAlexLichter/unovis/blob/26035c868a03a396de352a1de78bd9b7cdad5ec5/packages/ts/src/utils/text.ts#L372). So the original idea of making `fontSize` of `UnovisText` optional was probably the way to go. We just need to make sure that other functions that work with `UnovisText` don't rely on the provided `fontSize`.

Maybe it's worth renaming `UnovisText` to `UnovisTextPartial`, and adding a new `UnovisText` type with all the properties marked as mandatory.

lee00678:
Fixed!

PR #545, opened 2025-04-01
TS: Replacing deprecated JSX.Element with React.ReactNode

50rayn:
Noticed that #538 was replaced `JSX.Element` with `React.ReactNode`.  But the generated components from _ts_ folder still has `React.ReactNode`.

### DIFF
--- a/packages/react/autogen/component.ts
+++ b/packages/react/autogen/component.ts
@@ -23,7 +23,7 @@ export function getComponentCode (
     c.destroy()
   }`
   return `// !!! This code was automatically generated. You should not change it !!!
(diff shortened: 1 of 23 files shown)


lee00678:
@50rayn this update break the react build. See [here](https://github.com/f5/unovis/actions/runs/14203361271/job/40282749372?pr=545)

suryahanumandla:
@50rayn react build is still failing, please find more info [here](https://github.com/f5/unovis/actions/runs/14514694037/job/40860744407?pr=545)

50rayn:
Replaced JSX.Element with ReacElement.

PR #547, opened 2025-04-01
Component | Rolling Pin Legend

devgru:
My first attempt at introducing new component, legend with two labels and colored rects, to be used in Heatmap.
Screenshot 2025-04-01 at 23 13 16

I focused on exploring UnoVis codebase and was trying to follow the code style as closely as possible, but still most likely made a bunch of mistakes and false assumptions. I used barebone configuration and datatype for this component, which also might be debatable.

@rokotyan I'd be happy to get your feedback and any thoughts about what should be fixed/improved before we land this one.


devgru:
кусрусл

PR #552, opened 2025-04-11
Component | Chord: Fix link value zero issue

lee00678:
When specify link with 0 values, for example: `{ source: 'A', target: 'B', value: 0 }`, we get console errors due to partial `NaN` value in the path. The graph also renders a ribbon even though the value is 0 and nothing should be rendered.
Screenshot 2025-04-11 at 12 55 52 PM

This PR fix this issue by filtering out value 0 before setting layout data in `link`
Screenshot 2025-04-11 at 12 56 09 PM


lee00678:
cc @rokotyan

packages/ts/src/components/chord-diagram/index.ts
```diff
@@ -136,6 +137,7 @@ export class ChordDiagram<
       l.target._state.value = (l.target._state.value || 0) + getNumber(l, linkValue)
     })
 
+    links = links.filter(d => d._state.value !== 0)
```
suryahanumandla, line 140: @lee00678 what if value is null or undefined?. Is it a valid case to consider here?

suryahanumandla: I could see the issue when value is null or undefined
image

lee00678: Good point. I suppose we should catch that as well.

suryahanumandla:
Verified, LGTM!
image

suryahanumandla (approved):

PR #563, opened 2025-05-10
Fix: XYLabels not rendering issue

pipriles:
This PR resolves two issues:

- **_getDataToRender** – yRange now correctly derives from this.yScale.range() instead of this.xScale.range().

- **isNumberWithinRange** – Handle when range is inverted


pipriles:

rokotyan:
@pipriles Awesome, thanks for the fixes!

lee00678 (approved):
Thanks for the fix!

PR #565, opened 2025-05-13
Component | Scatter: Only resolve label overlap when needed

### LINKED ISSUE #564: Performance: don't call `collideLabels` if there are no labels
I noticed a performance issue with the scatter component in the case of frequent re-renders, such as brushing in another view. After investigating, it looks like the `collideLabels` is the culprit, which is executing even in the case that there are no labels on the scatter plot points.

Image

It may give us a big performance win if we can short-circuit that logic to do nothing if there are no labels configured on the plot. Worst case we still iterate over all the points, and just return early if there are no text nodes. Best case we use the config to detect if labels are present, and avoid calling `collideLabels` altogether in the case of no labels.

curran:
Closes #564 

**Technical note:** The check if (this.config.label) alone isn't sufficient because:                                                                                      

 * The label config could be defined, but it might return empty/null values for all data points. For example, if the label accessor function is defined as `d => d.label` but none of the data points actually have a label property or all have empty values.                                       
 * By checking if any point actually has a label `(pointDataFlat.some(d => d._point.label))`, we avoid unnecessary label overlap resolution calculations when no labels will be displayed, even though the label accessor is configured.

### DIFF
--- a/packages/ts/src/components/scatter/index.ts
+++ b/packages/ts/src/components/scatter/index.ts
@@ -147,7 +147,18 @@ export class Scatter<Datum> extends XYComponentCore<Datum, ScatterConfigInterfac
     removePoints(points.exit<ScatterPoint<Datum>>(), this.xScale, this.yScale, duration)
 
     // Take care of overlapping labels
-    this._resolveLabelOverlap()
+    if (this._hasLabels()) {
+      this._resolveLabelOverlap()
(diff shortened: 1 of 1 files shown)


rokotyan:
Nice

lee00678 (approved):
This is very nice.

PR #567, opened 2025-05-16
Misc | Dependencies: Update elkjs to 0.10.0

rokotyan:
Bumping up `elkjs` because the version we've been using have issues


lee00678 (approved):
LGTM

PR #568, opened 2025-05-16
Component | Graph: Adjust node positioning in ELK layout to center nodes based on width and height

rokotyan:
We were not centering the node positions coming from ELK which was wrong


huazeez (approved):

lee00678:
PR was reviewed.

Issue #574, opened 2025-05-28
single dot is being shown when linkFlow is true, source and target are same in graphs


suryahanumandla:
- When **linkFlow** is true and the given source and target are same then the link flow particle is being shown, ideally we shouldn't show the dot as source and target are same.

![Image](https://github.com/user-attachments/assets/87aed45f-743d-425b-89dd-379596dc2c68)

![Image](https://github.com/user-attachments/assets/805f2f97-2497-478d-bdbf-887190abab6e)

https://github.com/user-attachments/assets/7c6b73db-6a0a-4e26-87ed-2db86d1c92c2

rokotyan:
@suryahanumandla Good observation! Maybe we should filter out the links that have the same source and target since we don't really have a way to render those, and emit a warning to the console

PR #575, opened 2025-05-28
Draft: Component | Graph: Fix link flow group dot issue

suryahanumandla:
https://github.com/f5/unovis/issues/574


lee00678 (dismissed):
Nice catch and fix!

rokotyan:
@lee00678 @suryahanumandla I'll leave a few comments soon

rokotyan (changes requested):
I would do this on the data model level by filtering out the links, since we already filter out the links that point to missing nodes:

https://github.com/f5/unovis/blob/817adc5d7df405b5e8ff98b6287a09db165c0811/packages/ts/src/data-models/graph.ts#L70-L82

Since none of our components support self links, I think it's reasonable thing to do. @lee00678 What do you think?

lee00678:
> I would do this on the data model level by filtering out the links, since we already filter out the links that point to missing nodes:
> 
> https://github.com/f5/unovis/blob/817adc5d7df405b5e8ff98b6287a09db165c0811/packages/ts/src/data-models/graph.ts#L70-L82
> 
> Since none of our components support self links, I think it's reasonable thing to do. @lee00678 What do you think?

Good point. @suryahanumandla could you please update?

packages/ts/src/data-models/graph.ts
```diff
@@ -100,7 +100,40 @@ export class GraphDataModel<
     this._connectedNodes = without(nodes, ...this._nonConnectedNodes)
 
     this._nodes = nodes
-    this._links = links.filter(l => l.source && l.target)
+    this._links = links.filter(l => l.source && l.target && this.hasDistinctEndpoints(l))
```
rokotyan, line 103: Why not just do `l.source !== l.target`? This way you won't need to define extra helper functions in the code.

packages/ts/src/data-models/graph.ts
```diff
@@ -100,7 +100,40 @@ export class GraphDataModel<
...
+   *
+   * @param {OutLink} ol - The link object to check
+   * @returns {boolean} - Returns true if both source and target exist and are different,
+   * false if they are the same or either endpoint is missing
+   */
+  private hasDistinctEndpoints (ol: OutLink): boolean {
+    const sourceId = this.getSourceOrTargetId(ol.source)
+    const targetId = this.getSourceOrTargetId(ol.target)
+    if (sourceId === targetId) {
+      console.warn(`Link connects node to itself. Source and target IDs(${sourceId}) are identical.`)
```
rokotyan, line 132: We prefix warning messages with `Unovis | <> ...`. In this specific case it's `Unovis | Graph Data Model: ...`

rokotyan (changes requested):

packages/ts/src/data-models/graph.ts
rokotyan: @suryahanumandla Just above this we populate source and target with the actual node objects:

https://github.com/f5/unovis/blob/8cd7049347ef82170f61429bc8ab76de1d78c640/packages/ts/src/data-models/graph.ts#L70-L74

The type of `links` should be generic `OutLink extends GraphLinkCore<N, L>`, so you shouldn't need to do any type casting. But there can be some tings off with the types, I'm not 100% sure

suryahanumandla: But if you see GraphLinkCore definition where **L extends GraphInputLink**

`GraphLinkCore<N extends GraphInputNode, L extends GraphInputLink>`   
https://github.com/f5/unovis/blob/main/packages/ts/src/types/graph.ts#L34-L35

rokotyan: @suryahanumandla `GraphLinkCore` defines `source` and `target` to be the actual node objects, so there's no type conflict.

https://github.com/f5/unovis/blob/53601dcd854e79f3747c179084cb8f876a860219/packages/ts/src/types/graph.ts#L35-L36

I've just tried this on my end and didn't encounter any problems.
```ts
    this._links = links.filter(l => {
      if (l.source === l.target) {
        console.warn(`Unovis | Graph Data Model: Skipping link ${l._id} because it has the same source and target`)
        return false
      }

      return l.source && l.target
    })
  }
```

rokotyan (approved):
Looks good. The only minor thing for the future: we usually have separate commits for core and dev app changes. Here's a good commit history example https://github.com/f5/unovis/pull/578/commits

lee00678 (approved):
Thanks for the fix!

suryahanumandla:
> Looks good. The only minor thing for the future: we usually have separate commits for core and dev app changes. Here's a good commit history example https://github.com/f5/unovis/pull/578/commits

Thanks for the suggestion.

PR #578, opened 2025-06-03
Component | Graph: Configurable spacing for Parallel layouts

rokotyan:
- `layoutParallelGroupSpacing`: Controls spacing between main groups (columns/rows) in parallel layouts
Default: `undefined` (automatically calculated based on container size)
- `layoutParallelNodeSpacing`: Controls spacing between individual nodes in parallel layouts
Accepts number for uniform spacing or [horizontal, vertical] array for different x/y spacing
Default: `undefined` (dynamic spacing)
- `layoutParallelSubGroupSpacing`: Controls spacing between sub-groups within parallel layouts
Default: `40` pixels

### Dev example
https://github.com/user-attachments/assets/becf12b7-628b-4d0c-9d98-d0ecaeeb5bb2

### Docs update
SCR-20250603-jaxc


packages/angular/src/components/timeline/timeline.component.ts
lee00678, line 1: I'm guessing this is here because of the timeline component update and not related to the graph layout change?

rokotyan: Oops, missed that after my rebase. But it looks like the wrappers for Brush and Timeline were not updated for a while so we need to update them too.

lee00678 (approved):

PR #582, opened 2025-06-10
Core | Docs: Fix - make fontSize as optional in UnovisText type

suryahanumandla:


lee00678:
A fix for https://github.com/f5/unovis/pull/542

packages/ts/src/components/annotations/types.ts
```diff
@@ -1,9 +1,9 @@
 import { LengthUnit } from 'types/misc'
-import { UnovisText, UnovisTextOptions } from 'types/text'
+import { UnovisTextOptions, UnovisTextWithRequiredFontSize } from 'types/text'
 
 
 export type AnnotationItem = Omit<UnovisTextOptions, 'x'|'y'|'width'> & {
-  content: string | UnovisText | UnovisText[];
+  content: string | UnovisTextWithRequiredFontSize | UnovisTextWithRequiredFontSize[];
```
rokotyan, line 6: As far as I remember, the original point was that the `fontSize` should be optional here?

rokotyan (commented):
> P.S. I find the `Core | Docs` prefix a bit confusing because there are no "docs" in the core of the library (they are a part of Website). Essentially this is an update to TS types. 
> 
> I know our commitlint rules can be a bit unclear, and writing clear and concise commit messages is an art that is difficult to master. However, clear messages can really help when browsing through the history. So, maybe something like "_Core | Types | UnovisText: Make `fontSize` optional_" will do a better job.

What you can do @suryahanumandla is to have 2 separate commits, since you did update the docs as well. But if you don't feel like having a separate commit since the doc update is pretty small, one commit works too. The `|` here is not a combination, think of it as `"The key part you worked on" | Name of the component/gallery example: Brief message about it`. For this case it could be something like: `Component | Annotation: Update fontsize as optional` etc.

packages/ts/src/components/annotations/types.ts
lee00678: I think this is because the `text` in `getWrappedText` requires a fontSize, which is content here. But for annotation itself, fontSize is not longer required.

suryahanumandla: @rokotyan `fontSize` is required in `getWrappedText` and its subsequent methods that depend on it.

lee00678:
Fixed via #585

PR #580, opened 2025-06-06
Component | Treemap: New component

rokotyan:
Successor of #556

PR #584, opened 2025-06-13
Component | Treemap component

lee00678:
Successor of https://github.com/f5/unovis/pull/580

- [x] Fix search crash when no matching data
- [x]  add transitions
- [x]  Update Documentation
- [x]  Label occlusion issue
- [x] Fix data update, parent level font size issue

-- see how the parent level font size was set to large after search being reset:

https://github.com/user-attachments/assets/642116eb-f266-4d87-8cef-91fd36e4ec8b

-- fix:

https://github.com/user-attachments/assets/6e3c6a52-fc59-4e6c-a819-2456e2806483

https://github.com/user-attachments/assets/908e47ba-f613-40e0-b27f-091f1b80bd12


lee00678:
@rokotyan @curran made some update. The label issue was caused by `clipPath` calculation being inaccurate, only happens when custom value is passed in. I was not able to fix that with the time I have. I'm ok with merging this in, and have a follow up fix for that. Thoughts?

packages/ts/src/components/treemap/index.ts
```diff
@@ -0,0 +1,319 @@
...
+      .attr('x', d => d.x0)
+      .attr('y', d => d.y0)
+      .attr('width', d => d.x1 - d.x0)
+      .attr('height', d => d.y1 - d.y0)
+      .style('fill', d => d._fill ?? getColor(d, config.tileColor))
+      .style('opacity', 0)
+      .style('cursor', config.showTileClickAffordance ? d => !d.children ? 'pointer' : null : null)
+
+    tiles.merge(tilesEnter).select(`rect.${s.tile}`)
+      .transition()
```
rokotyan: Is there a reason for not using `smartTransition` here and in other places?

packages/ts/src/components/treemap/index.ts
```diff
@@ -0,0 +1,319 @@
...
+      .ease(easeCubicInOut)
+      .style('fill', d => d._fill ?? getColor(d, config.tileColor))
+      .style('opacity', 1)
+      .attr('x', d => d.x0)
+      .attr('y', d => d.y0)
+      .attr('width', d => d.x1 - d.x0)
+      .attr('height', d => d.y1 - d.y0)
+
+    // Update clipPath rects
+    tiles.merge(tilesEnter).select('clipPath rect')
```
rokotyan: I think there's no need to do `tiles.merge(tilesEnter)` multiple times, we can assign it to a variable and reuse.

It would be nice to use generics in the `select` calls, i.e. `tilesMerged.select<SVGRectElement>('clipPath rect')`

packages/ts/src/components/treemap/index.ts
```diff
@@ -0,0 +1,319 @@
...
+    mergedTiles.select(`text.${s.label}`)
+      .filter(d => {
+        const isAllowedNode = config.labelInternalNodes ? true : !d.children
+        return isAllowedNode && this.isTileLargeEnough(d)
+      })
+      .attr('clip-path', d => `url(#clip-${d._id})`)
+      .attr('x', d => d.x0 + config.labelOffsetX)
+      .attr('y', d => d.y0 + config.labelOffsetY)
+      .each(function (d) {
+        const text = select(this) as unknown as Selection<SVGTextElement, any, SVGElement, any>
```
rokotyan: Casting to `unknown` and then using `any` looks hacky. I'm pretty sure it can be avoided by using generics.
Also using an arrow function is preferred, i.e: 
```ts
.each((d, i, els) => {
  const text = select<SVGTextElement, TreemapNode<Datum>>(els[i])
  ...
```

packages/ts/src/components/treemap/index.ts
```diff
@@ -0,0 +1,319 @@
...
+
+        const label = `${d.data.key}: ${numberFormat(d.value)}`
+        text.text(label)
+
+        // Set text color based on background darkness
+        const backgroundColor = d._fill ?? getColor(d, config.tileColor)
+        const textColor = backgroundColor && isDarkBackground(backgroundColor) ? '#ffffff' : '#000000'
```
rokotyan: I think we need to store these colors as CSS variables and use them instead, this will allow the users to modify them.

packages/ts/src/components/treemap/style.ts
```diff
@@ -0,0 +1,71 @@
...
+
+// Utils
+import { getCssVarNames, injectGlobalCssVariables } from 'utils/style'
+
+const cssVarDefaults = {
+  '--vis-treemap-tile-stroke-color': '#fff',
+  '--vis-treemap-tile-stroke-width': '2px',
+  '--vis-treemap-tile-hover-stroke-color': '#fff',
+  '--vis-treemap-tile-fill-color': '#B9BEC3',
+  '--vis-treemap-tile-background-color': '#fff',
```
rokotyan, line 12: This variable doesn't have a dark version, maybe it should

packages/ts/src/components/treemap/style.ts
```diff
@@ -0,0 +1,71 @@
...
+  label: g-tiles;
+`
+
+export const tileGroup = css`
+  label: tile-group;
+`
+
+export const tile = css`
+  label: tile;
+  stroke: var(--vis-treemap-tile-hover-stroke-color);
```
rokotyan: This is not typesafe. Instead of using the variable names directly we should use it from `variables`: `var(${variables.treemapTileHoverStrokeColor});`

packages/ts/src/components/treemap/style.ts
```diff
@@ -0,0 +1,71 @@
...
+  stroke-opacity: 0;
+  transition: stroke-opacity 100ms ease-in-out;
```
rokotyan: @curran  Why do we need to animate stroke opacity? I'm concerned that this won't respect the duration specified in the chart config, also 100ms is really-really fast.

packages/ts/src/components/treemap/index.ts
```diff
@@ -0,0 +1,319 @@
...
+import { SeriesDataModel } from 'data-models/series'
+import { getColor, brighter, getHexValue, isDarkBackground } from 'utils/color'
+import { getString, getNumber, isNumber } from 'utils/data'
+import { TreemapConfigInterface, TreemapDefaultConfig } from './config'
+import { TreemapDatum, TreemapNode } from './types'
+
+import * as s from './style' // Minimum pixel size for showing labels
+
+
+const LIGHTNESS_VARIATION_AMOUNT = 0.08
```
rokotyan: Defining constants like this is not the bast practice. We should either make them a part of the config, or move to a dedicated `constants.ts` file and prefix with `TREEMAP_`

packages/ts/src/components/treemap/index.ts
```diff
@@ -0,0 +1,319 @@
...
+
+export class Treemap<Datum> extends ComponentCore<Datum[], TreemapConfigInterface<Datum>> {
+  static selectors = s
+  protected _defaultConfig = TreemapDefaultConfig as TreemapConfigInterface<Datum>
+  public config: TreemapConfigInterface<Datum> = this._defaultConfig
+
+  datamodel: SeriesDataModel<Datum> = new SeriesDataModel()
+  tiles: Selection<SVGGElement, unknown, SVGGElement, unknown>
+
+  private isTileLargeEnough (d: TreemapNode<Datum>): boolean {
```
rokotyan: We usually prefix private methods and variables with `_` to make it more obvious when you're reading the code

packages/ts/src/components/treemap/index.ts
```diff
@@ -0,0 +1,319 @@
...
+    const w = d.x1 - d.x0
+    const h = d.y1 - d.y0
+    return (w >= MIN_TILE_SIZE_FOR_LABEL) && (h >= MIN_TILE_SIZE_FOR_LABEL)
+  }
+
+  private getTileLightness (node: TreemapNode<Datum>, siblings: TreemapNode<Datum>[]): number {
```
rokotyan: same as above

rokotyan (commented):
I should have done this in-depth review earlier, sorry. Better late than never.

@curran I feel like the majority of my comments are addressed to you. You won't be able to update this PR, but please read them through and take them into account for future contributions.

@lee00678 There are also a couple of visual issues:

1. On Safari the labels are off. This is because Safari next `dominant-baseline` to be applied to `<tspan>` elements, not the top `<text>` element.
  image

2. The animations are not applied to the correct elements which probably means the ids are wrong. Can't attach a video now, will try to do it later.

curran:
Thank you so much for this detailed review!

packages/ts/src/components/treemap/style.ts
lee00678: Screenshot 2025-06-24 at 2 49 37 PM
I think this component doesn't require a dark mode.

packages/ts/src/components/treemap/index.ts
lee00678: Move `LIGHTNESS_VARIATION_AMOUNT ` and `MIN_TILE_SIZE_FOR_LABEL` to config file to make them configurable.

packages/ts/src/components/treemap/style.ts
lee00678: I've added a dark theme here. 
Screenshot 2025-07-01 at 11 25 24 AM

packages/ts/src/components/treemap/config.ts
```diff
@@ -0,0 +1,88 @@
...
+  tileLabelMediumFontSize?: number;
+
+  /** Large font size for leaf labels (used when enableTileLabelFontSizeVariation is true). Default: `24` */
+  tileLabelLargeFontSize?: number;
+
+
+  /** Flag for showing cursor:pointer to indicate leaf tiles are clickable. Default: `undefined` */
+  showTileClickAffordance?: boolean;
+
+  lightnessVariationAmount?: number;
```
rokotyan, line 64: That would be nice to add some jsdoc here

packages/ts/src/components/treemap/index.ts
```diff
@@ -0,0 +1,304 @@
...
+import { ComponentCore } from 'core/component'
+import { SeriesDataModel } from 'data-models/series'
+import { getColor, brighter, getHexValue } from 'utils/color'
+import { getString, getNumber, isNumber } from 'utils/data'
+import { smartTransition } from 'utils/d3'
+import { TreemapConfigInterface, TreemapDefaultConfig } from './config'
+import { TreemapDatum, TreemapNode } from './types'
+import * as s from './style'
+
+const defaultNumberFormat = (value: number): string => `${value}`
```
rokotyan: Maybe instead of having this as a global function, we can define it as a component method?

packages/ts/src/components/treemap/index.ts
```diff
@@ -0,0 +1,304 @@
...
+      .attr('width', (d: TreemapNode<Datum>) => Math.max(0.1, d.x1 - d.x0))
+      .attr('height', (d: TreemapNode<Datum>) => Math.max(0.1, d.y1 - d.y0))
+      .attr('rx', rx)
+      .attr('ry', rx)
+
+    clipPaths.exit().remove()
+
+    tilesEnter
+      .append('g')
+      .attr('class', 'label-group')
```
rokotyan: I think we need to define a dynamic class name for this in `style.ts` and use it

packages/ts/src/components/treemap/index.ts
```diff
@@ -0,0 +1,304 @@
...
+      .append('text')
+      .attr('class', s.label)
+      .style('opacity', 0)
+
+    mergedTiles.select('g.label-group text')
```
rokotyan: I believe that can simply use the `s.label` selector here set as a class name above

packages/ts/src/components/treemap/index.ts
```diff
@@ -0,0 +1,304 @@
...
+      .text(d => `${d.data.key}: ${numberFormat(d.value)}`)
```
rokotyan: That would be great to give users a way to configure it via config

packages/ts/src/components/treemap/index.ts
```diff
@@ -0,0 +1,304 @@
...
+      .style('font-size', function (d) {
+        const sqrtVal = Math.sqrt(d.value ?? 0)
+        return config.enableTileLabelFontSizeVariation && !d.children
+          ? `${fontSizeScale(sqrtVal)}px`
+          : `${fontSizeScale.range()[1]}px`
+      })
+      .attr('dominant-baseline', 'middle')
+      .each((_, i, nodes) => {
+        const text = select(nodes[i] as SVGTextElement)
+        text.selectAll('tspan').attr('dominant-baseline', 'middle')
```
rokotyan: Does this need a loop? You can probably simply do `selectAll` without it

packages/ts/src/components/treemap/index.ts
```diff
@@ -0,0 +1,304 @@
...
+      .attr('ry', rx)
+
+    clipPaths.exit().remove()
+
+    tilesEnter
+      .append('g')
+      .attr('class', 'label-group')
+      .append('text')
+      .attr('class', s.label)
+      .style('opacity', 0)
```
rokotyan, line 264: I think we need to set the initial x and y values here, otherwise the label transition looks off      

https://github.com/user-attachments/assets/b0be8d97-5d19-4faa-b3a1-abc97073a953

packages/ts/src/components/treemap/index.ts
```diff
@@ -0,0 +1,304 @@
...
+      .style('opacity', 1)
+
+    // Hide labels that don't meet criteria
+    mergedTiles.select(`text.${s.label}`)
+      .style('display', d => {
+        const isAllowedNode = config.labelInternalNodes ? true : !d.children
+        return isAllowedNode && this._isTileLargeEnough(d) ? null : 'none'
+      })
+      // Make the internal labels semibold
+      .style('font-weight', d => d.children ? '500' : 'normal')
```
rokotyan: I would do this via assigning a class name and having a CSS configurable variable

packages/ts/src/components/treemap/style.ts
```diff
@@ -0,0 +1,69 @@
...
+  label: tile-group;
+`
+
+export const tile = css`
+  label: tile;
+  stroke: var(--vis-treemap-tile-hover-stroke-color);
+  stroke-opacity: 0;
+
+  &:hover {
+    stroke-opacity: 1;
```
rokotyan: Not everyone needs interactivity, so I would add a CSS variable for this and made it 0 by default to avoid the unnecessary highlight:

https://github.com/user-attachments/assets/01d44fd7-4a61-4134-8a4d-a25b9de1ceca

packages/dev/src/examples/misc/treemap/search/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+// Number formatter for population values,
+// using B for billions and M for millions.
+const populationFormatRaw = format('.3s')
+const populationFormat = (value: number): string => populationFormatRaw(value)
+  .replace('G', 'B')
+
+const getTooltipContent = (d: TreemapNode<TreemapExampleDatum>): string => {
+  const isLeafNode = d.data.datum !== undefined
+  return isLeafNode ? `
+    <div class="${s.tooltipTitle}">${d.data.datum.name}</div>
```
rokotyan: I see some type errors in this example that would be great to resolve.
Image

Also, managing color persistence is currently outside the library's responsibility, but it would be nice to show in this example how to have persistent colors. Currently when you search, the color of the tiles changes

https://github.com/user-attachments/assets/1a10071b-512a-4a86-96b5-acc694dcc43d

rokotyan (commented):
@lee00678 This is great progress, thanks!

I've left a few more comments and suggestions to the PR. Happy to discuss them.

cc @curran

lee00678:
https://github.com/user-attachments/assets/aea00a6d-52f6-4197-a01b-ed4ffb84ee75

Fixed text init position issue, turned off highlight by default, assigning tileColor so we can preserve the color.

curran:
LGTM!

IARI:
I've been watching the evolution of the issue ond the PRs for a while, and am looking forward to this :)
+1 - and thanks a lot to curran for the work!

suryahanumandla:
LGTM!

suryahanumandla (approved):
LGTM!

curran:
Woohoo! Finally. Thank you all for the collective effort here to clean this up. The power of Open Source!

PR #585, opened 2025-06-16
Core | Types | UnovisText: Making `fontSize` optional

rokotyan:
Successor of #582

I think that making `fontSize` in `UnovisText` optional, and not requiring it in any of the text function, is a more elegant way to solve the problem.

### DIFF
--- a/packages/ts/src/utils/text.ts
+++ b/packages/ts/src/utils/text.ts
@@ -225,7 +225,7 @@ export function estimateStringPixelLength (
  * @param {(string | number)} [fontSize] - The font size of the string.
  * @returns {number} The precise length of the string in pixels.
  */
-export function getPreciseStringLengthPx (str: string, fontFamily?: string, fontSize?: string | number): number {
+export function getPreciseStringLengthPx (str: string, fontFamily: string, fontSize: string | number): number {
   const svgNS = 'http://www.w3.org/2000/svg'
   const svg = document.createElementNS(svgNS, 'svg')
   const text = document.createElementNS(svgNS, 'text')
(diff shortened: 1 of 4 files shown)


lee00678:
also linking https://github.com/f5/unovis/pull/542

suryahanumandla:
LGTM!

packages/ts/src/utils/text.ts
```diff
@@ -300,6 +300,9 @@ function breakTextIntoLines (
 ): string[] {
   const text = `${textBlock.text}`
   if (!text) return []
+  const fontSize = textBlock.fontSize ?? UNOVIS_TEXT_DEFAULT.fontSize
+  const fontFamily = textBlock.fontFamily ?? UNOVIS_TEXT_DEFAULT.fontFamily
+  const fontWidthToHeightRatio: number | undefined = textBlock.fontWidthToHeightRatio ?? UNOVIS_TEXT_DEFAULT.fontWidthToHeightRatio
```
suryahanumandla, line 305: @rokotyan I haven't find any default value set part of `UNOVIS_TEXT_DEFAULT.fontWidthToHeightRatio`. https://github.com/f5/unovis/blob/7e0ed7a874ed9f49370c3be99d81cfaec0020aee/packages/ts/src/styles/index.ts#L11-L18
I can see one here https://github.com/f5/unovis/blob/7e0ed7a874ed9f49370c3be99d81cfaec0020aee/packages/ts/src/styles/index.ts#L8 but we haven't asssigned to `fontWidthToHeightRatio`

rokotyan: As far as I remember we keep it `undefined` in `UNOVIS_TEXT_DEFAULT` to allow a fallback to `UNOVIS_FONT_WH_RATIO_DEFAULT` in some of the text utils. Maybe it's not really needed, but it needs an investigation first. I suggest to address that later

suryahanumandla (approved):

PR #590, opened 2025-06-25
Misc: Migrate @unovis/ts build to Vite

rokotyan:
Another attempt to migrate the main build to Vite. Will test first and then mark as ready to review.

Successor of #495


50rayn:
It should be taken into consideration that the recently released v7. For future upgrade possibilities, the breaking changes should be followed ([link](https://vite.dev/guide/migration.html#migration-from-v6)).

PR #591, opened 2025-06-27
Component | BulletLegend: Support multiple colors per item

curran:


curran:
Current status:

 * It's basically working, but the changes break the existing legend examples, so it's not ready
 * More work is required to find the minimal set of changes that will maintain the working state of all existing examples, and get the new one working

New example working (need to iterate on the spacing, but basically it's working):

image

Old examples broken:

image

image

packages/ts/src/components/bullet-legend/types.ts
```diff
@@ -4,6 +4,7 @@ import { SymbolType } from 'types/symbol'
 export interface BulletLegendItemInterface {
   name: string | number;
   color?: string;
+  colors?: Array<string>;
```
rokotyan: I think we can use the existing color property by allowing it to accept an array:
```ts
color?: string | string[];
```

packages/ts/src/components/bullet-legend/style.ts
```diff
@@ -67,9 +67,10 @@ export const label = css`
 export const bullet = css`
   flex: 0 0 auto;
   label: legendItemBullet;
+  display: inline-block;
+  vertical-align: middle;
   margin-right: var(--vis-legend-bullet-label-spacing);
   height: var(--vis-legend-bullet-size);
-  width: var(--vis-legend-bullet-size);
```
rokotyan: Is there a reason to remove `width` from the style and add new properties?

packages/ts/src/components/bullet-legend/modules/shape.ts
```diff
@@ -43,59 +43,85 @@ export function createBullets (
...
-        .style('opacity', opacity)
-        .style('stroke-width', '3px')
-        .style('fill', null)
-        .style('fill-opacity', null)
-        .style('marker-start', 'none')
-        .style('marker-end', 'none')
+    const svgsMerged = svgs.merge(svgsEnter)
+
+    if (d.colors) {
+      bulletContainer.style('display', 'flex')
```
rokotyan: The approach of modifying the container from children can be confusing. I think we should get rid of this `if (d.colors) {} else {}` statement and set the styles at the top level, ideally in the `style.ts` file

packages/ts/src/components/bullet-legend/style.ts
curran: It's WIP, there was an issue with the circles shrinking to fit the container, and I thought this would help. We'll need to keep this width somewhere, but maybe on a nested element within this container, otherwise the circles shrink (e.g. if there are 3 in one item, they all shrink so their combined radius matches the `width` here. This change is probably what's breaking the original examples. Will try to find a way!

curran:
Current status:

 * It's working well

image

 * The width as a string was somewhat tricky to deal with, since we need to parse it and do some math on it, but otherwise I think this is good to go!
 * The existing examples all work again
 * I've been force-pushing and using `git commit --amend`, since that seems to be the desired style with this repo (but we can't see the intermediate commits - I guess that's fine)
 * Requesting re-review @rokotyan . Thanks!

packages/ts/src/components/bullet-legend/modules/shape.ts
```diff
@@ -50,52 +51,68 @@ export function updateBullets (
   container.each((d, i, els) => {
     const shape = getString(d, config.bulletShape, i) as BulletShape
     const color = getColor(d, colorAccessor, i)
```
rokotyan: Looks like there's no need to create this variable and call `getColor` if `d.color` is defined

packages/ts/src/components/bullet-legend/modules/shape.ts
```diff
@@ -50,52 +51,68 @@ export function updateBullets (
...
-    const width = BULLET_SIZE
+    const colorValue = d.color ?? color
+    const colors = Array.isArray(colorValue) ? colorValue : [colorValue ?? '#000']
```
rokotyan: Use can use the `ensureArray` function here.

Also I'm not sure about the fallback to `#000`, I think the color will be automatically defined by CSS if the color value is missing, what do you think?

packages/ts/src/components/bullet-legend/constants.ts
```diff
@@ -0,0 +1,2 @@
+// Default bullet size in pixels (used as fallback when bulletSize is null)
+export const DEFAULT_BULLET_SIZE = '20'
```
rokotyan: Is there a reason to provide it as a string here? How does it correspond to the `--vis-legend-bullet-size` CSS variable?

packages/ts/src/components/bullet-legend/index.ts
```diff
@@ -66,7 +69,13 @@ export class BulletLegend {
       .call(createBullets)
 
     legendItemsMerged.select<SVGElement>(`.${s.bullet}`)
-      .style('width', config.bulletSize)
+      .style('width', (d: BulletLegendItemInterface) => {
+        const colors = Array.isArray(d.color) ? d.color : [d.color]
+        const numColors = colors.length
+        const baseSize = parseInt(config.bulletSize || DEFAULT_BULLET_SIZE)
```
rokotyan: What if `config.bulletSize` has units like `pt` or `em`? It's better to use the `to-px` package to convert

packages/ts/src/components/bullet-legend/constants.ts
curran: The existing type had it declared as string.

```ts
  /** Bullet shape size, mapped to the width and height CSS properties. Default: `null` */
  bulletSize?: string | null;
```

in 

```
packages/ts/src/components/bullet-legend/config.ts
```

rokotyan: The `bulletSize` config property is exposed to the users of the library to allow them to specify the bullet size with the unit of their choice. And `DEFAULT_BULLET_SIZE` is an internal constant that no user will ever redefine. Or am I missing something?

curran: Great point! Will update that to be a number.

packages/ts/src/components/bullet-legend/types.ts
curran: Done.

packages/ts/src/components/bullet-legend/style.ts
curran: Done.

packages/ts/src/components/bullet-legend/modules/shape.ts
curran: It's the default colors that were proving problematic earlier. I don't fully understand how they work, but the `getColor` seems to handle the case properly.

packages/ts/src/components/bullet-legend/constants.ts
curran: Done.

packages/ts/src/components/bullet-legend/index.ts
curran: Done.

packages/ts/src/components/bullet-legend/modules/shape.ts
curran: But it does the right thing when `d.color` is _not_ defined also, which I'm not sure how to handle differently from this. Did you have a concrete change in mind?

curran: OK I cleaned it up to this:

```js
const colors = ensureArray(d.color ?? getColor(d, colorAccessor, i))
```

I think this feedback is addressed.

packages/ts/src/components/bullet-legend/modules/shape.ts
curran: OK I cleaned it up to this:

```js
const colors = ensureArray(d.color ?? getColor(d, colorAccessor, i))
```

I think this feedback is addressed.

curran:
Thanks for reviewing @rokotyan ! I've addressed your points of feedback. Requesting re-review please. Thank you.

rokotyan:
Thanks for the tweaks @curran! There's one question that remains unanswered: How does `DEFAULT_BULLET_SIZE` correspond to the `--vis-legend-bullet-size` CSS variable?

curran:
Yeah one is 9 and one is 20. I suppose we should have consistency between them. I'll change the 20 to 9 for now.

But that brings up the larger question of: why do we need to define the constant in both CSS and JS? Could that be avoided to have just one centralized definition?

Is the idea that users of the Unovis library can specify the value using CSS variables? In that case we would want to "read" the value of the CSS variable at runtime, right?

Or would it make sense to remove that CSS var?

Thanks for the feedback! Almost there with this one.

curran:
I guess something like this would be ideal:

- CSS variables remain the standard way to handle theming/customization
- Keeps styling concerns in CSS
- Allows users to override the value with CSS without touching JavaScript
- Maintains existing API where `bulletSize: null` means "use the default"
- Eliminates duplication by making CSS the single source of truth

**Implementation:**
1. Remove `DEFAULT_BULLET_SIZE` constant from constants.ts
2. Update JavaScript code to read CSS variable value when `config.bulletSize` is null
3. Use `getComputedStyle(document.documentElement).getPropertyValue('--vis-legend-bullet-size')`

curran:
image


OK I made one final change - remove the JS constant and instead read the value from the CSS variable. Also changed the example so that it does actually fall back to the default, so it tests this code path. I think it's ready to go!

packages/ts/src/components/bullet-legend/config.ts
```diff
@@ -26,6 +26,8 @@ export interface BulletLegendConfigInterface {
   labelMaxWidth?: string | null;
   /** Bullet shape size, mapped to the width and height CSS properties. Default: `null` */
   bulletSize?: string | null;
+  /** Spacing between multiple bullet symbols in pixels. Default: `3` */
```
rokotyan: The default value below is actually `4` not `3`

packages/ts/src/components/bullet-legend/modules/shape.ts
```diff
@@ -49,53 +50,68 @@ export function updateBullets (
...
   container.each((d, i, els) => {
     const shape = getString(d, config.bulletShape, i) as BulletShape
-    const color = getColor(d, colorAccessor, i)
-    const width = BULLET_SIZE
+    const colors = ensureArray(d.color ?? getColor(d, colorAccessor, i))
+    const numBullets = colors.length
+    const bulletWidth = BULLET_SIZE
+    const defaultBulletSize = toPx(getComputedStyle(els[i]).getPropertyValue('--vis-legend-bullet-size'))
+    const spacing = config.bulletSpacing * (BULLET_SIZE / defaultBulletSize) // Scale spacing relative to bullet size
+    const width = bulletWidth * numBullets + spacing * (numBullets - 1)
```
rokotyan: This looks like a duplicate code, or it's very similar to the one from the main index file. Can we separate it into a function?

curran: Done

curran:
Address points of feedback:

 * Introduced `getBulletsTotalWidth`
 * Fixed inconsistency between 3 and 4

Ready for final final review @rokotyan ! Thank you.

lee00678 (approved):

PR #593, opened 2025-06-27
Component | Crosshair: Add synchronized function

lee00678:
This PR add the synchronized function to crosshair to enable it between multiple charts.

- [ ] Add a dev example.
- [ ] Add Doc
- [ ] Add website example

https://github.com/user-attachments/assets/d5af4922-74f8-4b96-812a-12a237bbe1ab


packages/ts/src/components/crosshair/config.ts
```diff
@@ -44,6 +44,16 @@ export interface CrosshairConfigInterface<Datum> extends WithOptional<XYComponen
    * Default: `undefined`
   */
   getCircles?: (x: number | Date | Date, data: Datum[], yScale: ContinuousScale) => CrosshairCircle[];
+  /** X position in data space (not pixels) for synchronized crosshair. Default: `undefined` */
+  xPosition?: number | Date;
+  /** Enable synchronized crosshair across multiple charts. Default: `false` */
+  enableSync?: boolean;
```
rokotyan: This `enableSync` seems redundant to me — we already have `xPosition`, `forceShow`, `isActive`. Why do we need so many?

suryahanumandla:
@lee00678 couple of observations

- [ ]  React gallery example: showing extra tooltip

     image

- [ ] typescript gallery example: only one tooltip being shown
     image

- [ ] typescript gallery example: got broken when we do zoomin or zoomout
     
     https://github.com/user-attachments/assets/7bac3764-43bf-4730-80aa-a9fecf33449c

-[ ] Is it a valid use case? I see tooltip moves dynamically along with the cursor in one chart while other tooltips remain fixed at their respective data points until the cursor reaches the next data point

   https://github.com/user-attachments/assets/6c47e990-f066-4fdc-bc3e-018c35c2b6a4

lee00678:
Close in favor of https://github.com/f5/unovis/pull/595

PR #595, opened 2025-07-03
Component | Crosshair: Add Synchronization

lee00678:
This PR uses a different approach than the previous one. We are in favor of not allowing component communicating to each other. 

- [x] Add Gallery example
- [x] Add Dev example
- [x] Add Documentation


packages/ts/src/components/crosshair/index.ts
```diff
@@ -162,25 +205,77 @@ export class Crosshair<Datum> extends XYComponentCore<Datum, CrosshairConfigInte
...
+    const tooltip = this.config.tooltip ?? this.tooltip
+
     if (!tooltip) return
 
+    const { isActive, enableSync } = this._getSyncState()
+
+    // For synchronized charts, we need to find the datum if it doesn't exist
+    let datum = this.datum
+    if (!datum && enableSync && !isActive && this._currentXData !== undefined) {
+      if (typeof this._currentXData === 'number') {
```
rokotyan: Can use `isNumber`

packages/ts/src/components/crosshair/index.ts
```diff
@@ -162,25 +205,77 @@ export class Crosshair<Datum> extends XYComponentCore<Datum, CrosshairConfigInte
     else this._hideTooltip()
   }
 
+  _setupDocumentMouseTracking (): void {
```
rokotyan: Why not to handle this inside existing `_onMouseMove`?

packages/ts/src/components/crosshair/index.ts
```diff
@@ -34,6 +34,9 @@ export class Crosshair<Datum> extends XYComponentCore<Datum, CrosshairConfigInte
   datumIndex: number
   show = false
   private _animFrameId: number = null
+  private _currentXData: number | Date | undefined = undefined
```
rokotyan: When I read `this._currentXData` I expect it to have type `Datum`. Maybe `this._dataX` since we already have `this.x`?

packages/ts/src/components/crosshair/index.ts
```diff
@@ -50,44 +53,47 @@ export class Crosshair<Datum> extends XYComponentCore<Datum, CrosshairConfigInte
...
     this.container.on('mousemove.crosshair', this._onMouseMove.bind(this))
     this.container.on('mouseout.crosshair', this._onMouseOut.bind(this))
+    this._render()
   }
 
   _render (customDuration?: number): void {
     const { config } = this
-    if (config.snapToData && !this.datum) return
     const duration = isNumber(customDuration) ? customDuration : config.duration
+    const { isActive, enableSync } = this._getSyncState()
```
rokotyan, line 93: I'm not sure it's worth introducing the whole concept of `sync` into the component because, logically, all the syncing is happening outside of it. What we do on the Crosshair component is simply enforce it to be displayed at a specific position.

packages/ts/src/components/crosshair/index.ts
```diff
@@ -50,44 +52,55 @@ export class Crosshair<Datum> extends XYComponentCore<Datum, CrosshairConfigInte
...
+    }
 
-    smartTransition(this.g, duration)
-      .style('opacity', this.show ? 1 : 0)
+    smartTransition(this.g, duration).style('opacity', this.show ? 1 : 0)
 
     this.line
       .attr('y1', 0)
-      .attr('y1', this._height)
+      .attr('y2', this._height)
```
suryahanumandla, line 103: any reason for changing it to 'y2'?

packages/ts/src/components/crosshair/index.ts
```diff
@@ -124,17 +137,55 @@ export class Crosshair<Datum> extends XYComponentCore<Datum, CrosshairConfigInte
     this._onMouseOut()
   }
 
+  destroy (): void {
```
suryahanumandla, line 140: please add jsdoc comment for all the new introduced methods

suryahanumandla:
@lee00678 couple of observations

-  X-axis is not visible in typescript gallery example

image

-  typescript gallery example got broken when we do zoomin or zoomout


https://github.com/user-attachments/assets/b630a291-5089-4b4e-8bb3-84d607ff9b63

- React gallery example, when we move cursor little fast the tooltip is being hidden is it expected behavior?


https://github.com/user-attachments/assets/7ce55cd1-1c14-4fc6-be31-0ac4bec46e01

- Tooltip is being disappeared in solid gallery example 

https://github.com/user-attachments/assets/f8b8b655-d970-45bf-b154-b8500ecadcb7

- Tooltip is being disappeared in svelte gallery example 

https://github.com/user-attachments/assets/db1e1118-019f-430d-88b5-2d8fde26cb0e

packages/ts/src/components/crosshair/index.ts
rokotyan: y2 seems to be correct

packages/ts/src/components/crosshair/index.ts
rokotyan: Let's discuss this a little. Why do you think it's a good practice?

When there are too many JSDoc comments in the code, I find them more distracting than helpful, especially in the age of AI, which can easily explain what happens in your codebase at a high level. 

I personally prefer to have comments only in places where they really matter and explain something that is not obvious. This "destroy" method, however, is not very sophisticated and is mainly for internal use.

suryahanumandla: > I personally prefer to have comments only in places where they really matter and explain something that is not obvious. This "destroy" method, however, is not very sophisticated and is mainly for internal use.

I just added comment at destroy method but intention was to include at all newly introduced methods. 

I do agree AI tools can explain code behavior, but JSDoc gives developers immediate visibility into the purpose of each code element without needing to query or dig deeper.

lee00678:
> @lee00678 couple of observations
> 
> * X-axis is not visible in typescript gallery example

 This is happening for all ts examples. I think we should address it in a separate PR.
> 
> * typescript gallery example got broken when we do zoomin or zoomout

fixed

packages/ts/src/components/crosshair/index.ts
lee00678: > I personally prefer to have comments only in places where they really matter and explain something that is not obvious. This "destroy" method, however, is not very sophisticated and is mainly for internal use.

I agree with this. I found too much comments distracting. I think it's personal preference. @suryahanumandla if you think any update unclear, let me know I'm happy to add some simple JSDoc.

lee00678:
close in favor of https://github.com/f5/unovis/pull/598

PR #609, opened 2025-07-23
Component | Axis: Label trim and wrap

lee00678:
This PR fix long label rendering issue:
Screenshot_2025-07-03_at_9 39 35_AM

Adding `labelTextTrimType` and `labelTextTrimType`.
Screenshot 2025-07-23 at 3 04 53 PM


rokotyan:
Great addition! It would be nice test how well the axis transitions and label changes work.

packages/ts/src/components/axis/index.ts
```diff
@@ -80,35 +90,140 @@ export class Axis<Datum> extends XYComponentCore<Datum, AxisConfigInterface<Datu
...
+        const bleedY = this._axisRawBBox.height > this._height ? (this._axisRawBBox.height - this._height) / 2 : 0
         const top = bleedY
         const bottom = bleedY
 
+        // For Y-axis, calculate margin based on the actual rendered axis size
+        // This includes ticks, tick labels, and any padding
+        const axisWidth = this._axisRawBBox.width
+
+        // Add extra padding to ensure no overlap
+        const extraPadding = 12 // Increased padding for safety
```
rokotyan: @lee00678 Could you tell more about this overlap? Maybe it can be avoided by changing the text baseline?

packages/ts/src/components/axis/index.ts
```diff
@@ -362,33 +481,86 @@ export class Axis<Datum> extends XYComponentCore<Datum, AxisConfigInterface<Datu
...
+      labelHeight = labelBBox.height
+    } else {
+      const trimWidth = type === AxisType.X ? labelWidth : labelHeight
+      trimSVGText(
+        textElement,
+        trimWidth,
+        this.config.labelTextTrimType as TrimMode,
+        true,
+        parseInt(labelFontSize || '12', 10),
+        0.58
```
rokotyan: I don't think we should hardcode the width to hight ratio. Users might use different fonts

packages/dev/src/examples/xy-components/axis/axis-label/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+    (d: XYDataRecord) => d.y1,
+    (d: XYDataRecord) => d.y2,
+    () => Math.random(),
+    () => Math.random(),
+  ]
+  return (
+    <>
+      <VisXYContainer<XYDataRecord> data={generateXYDataRecords(15)} width={600}>
+        <VisLine x={d => d.x} y={accessors} duration={props.duration}/>
+        <VisAxis type='x' numTicks={15} tickFormat={(x: number) => `${x}`} duration={props.duration} label='S little longer X Axis X Axis supe r super suepra lskdjgaljgal fjglaijhlaidaij ijsdjflak  sladkjfalfgjlfijg jhglijhlij' labelTextFitMode={FitMode.Trim} position='bottom' />
```
rokotyan: Knowing that we're going to publish our examples publicly in the future, that would be nice to have more meaningful text here.

packages/ts/src/components/axis/index.ts
```diff
@@ -362,33 +360,82 @@ export class Axis<Datum> extends XYComponentCore<Datum, AxisConfigInterface<Datu
...
+    // Calculate offset after wrapping to get accurate dimensions
+    let labelWidth = axisWidth
+    let labelHeight = axisHeight
+    if (labelTextFitMode === FitMode.Wrap) {
+      const labelBBox = textElement.node().getBBox()
+      labelWidth = labelBBox.width
+      labelHeight = labelBBox.height
+    } else {
+      const trimWidth = type === AxisType.X ? labelWidth : labelHeight
+      const styleDeclaration = getComputedStyle(textElement.node())
```
rokotyan, line 403: Theoretically I can see this a performance bottleneck as `getComputedStyle` causes a [reflow](https://gist.github.com/paulirish/5d52fb081b3570c81e3a). It can become a problem when we need axis to update frequently, e.g. when using a Brush.


We can to optimize in the future or now, @lee00678 up to you

rokotyan (approved):
This is great!, posted a coupel comments

packages/ts/src/components/axis/index.ts
lee00678: https://github.com/user-attachments/assets/2226d5a0-0b61-4a0c-a5c7-a925f9b2cc8f

Is this what you are referring to? It seems it's working well for now.

lee00678: https://github.com/user-attachments/assets/2226d5a0-0b61-4a0c-a5c7-a925f9b2cc8f

Is this what you are referring to? It seems it's working well for now.

rokotyan: Yes, maybe it's not that critical here. Just wanted to bring this up because it some situations it makes a big difference. Thanks for checking

suryahanumandla:
@lee00678 @rokotyan IMO, we should have a tooltip for trimmed label. WDYT?

lee00678:
> @lee00678 @rokotyan IMO, we should have a tooltip for trimmed label. WDYT?

I think that would be a good idea, but outside of the scope of this ticket.

suryahanumandla:
@lee00678 Y-label is little shaky other than that everything is looking great! 

https://github.com/user-attachments/assets/fd5b7d7e-e869-4fd9-abdc-39c7ada0ad7a

lee00678:
> @lee00678 Y-label is little shaky other than that everything is looking great!
> 
>  Screen.Recording.2025-08-11.at.7.34.31.PM.mov

I think this is ok since Axis gets re-rendered. Similar behavior happens on ticks as well.

rokotyan:
> > @lee00678 @rokotyan IMO, we should have a tooltip for trimmed label. WDYT?
> 
> I think that would be a good idea, but outside of the scope of this ticket.

We can consider using the built-in HTML capabilities for that, e.g. adding a `<title>` element inside `<text>` with the full label text. I wouldn't do more than that at the moment.

suryahanumandla (approved):
LGTM!

PR #611, opened 2025-07-30
Component | Stacked Bar: Add stacked spacing option

dennisadriaans:
This adds an option to apply spacing between stacked items within a single bar. Applies radius application across all stacked items when a chart-wide radius is set.

### DIFF
--- a/packages/website/docs/xy-charts/StackedBar.mdx
+++ b/packages/website/docs/xy-charts/StackedBar.mdx
@@ -26,6 +26,10 @@ The _Stacked Bar_ component has been designed to work together with _XY Containe
 _Stacked Bar_ can accept an array of y accessors to display values as a stack of bars:
 <XYWrapper {...stackedBarProps()} showContext="full" y={[d => d.y, d => d.y1, d => d.y2]}/>
 
+### Stacked Bar with Spacing `1.6`
+_Stacked Bar_ supports spacing between each stacked bars using the `stackSpacing` property. 
+<XYWrapper {...stackedBarProps()} showContext="full" y={[d => d.y, d => d.y1, d => d.y2]} stackSpacing={3}/>
+
 ## Orientation
 _Stacked Bar_ supports horizontal and vertical orientation.
 <XYWrapper {...stackedBarProps()} y={[d => d.y, d => d.y1, d => d.y2]} height={250} showAxes property="orientation" inputType="select" options={["horizontal","vertical"]}/>


packages/ts/src/components/stacked-bar/index.ts
```diff
@@ -220,33 +220,50 @@ export class StackedBar<Datum> extends XYComponentCore<Datum, StackedBarConfigIn
...
     const width = barWidth
 
+    // --- Spacing between stacked items ---
+    const stackSpacing = config.stackSpacing || 0
+    let hWithSpacing = h
+    let yWithSpacing = y
+    // Only apply spacing between inner segments (not at base or top)
+    const isInnerStack = accessorIndex > 0 && accessorIndex < this.getAccessors().length
+    if (!isEntering && stackSpacing > 0 && isInnerStack) {
+      if (this.isVertical()) {
```
lee00678: When there are both negative and positive value in the same group, 
Screenshot 2025-07-30 at 1 27 44 PM

dennisadriaans:
Spacing, radius and crosshair should be good for both positive and negative values now. The crosshair for horizontal orientation seems bugged, but it looks like this happens in the main branch too

lee00678:
> Spacing, radius and crosshair should be good for both positive and negative values now. The crosshair for horizontal orientation seems bugged, but it looks like this happens in the main branch too

Yea, we don't currently support horizontal crosshair.

lee00678:
@dennisadriaans this looks good! It should be nice to have some documentation for this on the website. Do you have bandwidth? If not, I can add it.

lee00678:
Added doc. 
Screenshot 2025-08-06 at 12 32 57 PM

lee00678 (approved):
Thanks for the contribution!

suryahanumandla:
- I see spacing is not same for all

image


@lee00678 @rokotyan 
what would be the expected behavior when user wants to have `roundedCorners` along with `stackSpacing`?

image

rokotyan:
@suryahanumandla Great observation about the spacing, thanks for double checking! Also the space at the bottom looks off.
 
image



> what would be the expected behavior when user wants to have roundedCorners along with stackSpacing?

I think the screenshot you've shared looks good. What is your concern?

dennisadriaans:
The inconsistent spacing happens when the stackedSpacing is increased. Currently, we subtract the spacing visually from a stacked item to keep it aligned with the value on the y-axis. I'm not sure how to correct this. Any suggestions?

suryahanumandla:
@dennisadriaans I'll take a look at it, thanks.

lee00678:
I'm converting this to draft until it's ready. cc @dennisadriaans

Issue #620, opened 2025-08-19
Axis: When `minMaxTicksOnly` is true there are no grid lines


rokotyan:
That would be great to have an option to enable grid lines when when `minMaxTicksOnly` is true.

PR #621, opened 2025-08-19
Axis: Show grid when `minMaxTicksOnly` is `true`

rokotyan:
Adding a new config property to show grid when  `minMaxTicksOnly` is set to `true` #620 
  
image

SCR-20250819-lnid


rokotyan:
Pushed one commit that adds a grid line to the max value, I think it looks better this way. Also refactored the code a little.

image

rokotyan:
Pushed another small tweak
image

PR #622, opened 2025-08-19
Timeline: Configurable positioning of small rounded segments

rokotyan:
There can be situations where you might not want to center small segments. Adding a config property to control this: `showEmptySegmentsCorrectPosition` (`true` by default). 

SCR-20250819-ncxn

https://github.com/user-attachments/assets/b86c1639-4abd-44d3-a6e8-bc738b54df2e


suryahanumandla:
LGTM!

image

PR #624, opened 2025-08-21
Component | Annotation: Add calc() support to annotation position

lee00678:
There is a need to have the annotation to always display at a certain location of the chart (right top corner, i.e.) when you don't have access to the chart's width / height. 

This PR adds support for `calc()` like what css has. `Min` in this case will always stay in the top right corner.
This implementation supports `calc(100% - (20px + 30px + 10px))`

https://github.com/user-attachments/assets/b28c1cc3-0a84-41ec-83d2-5b50fc4f0dd7


suryahanumandla:
@lee00678 with incorrect expression page is unresponsive

`x: 'calc(((100% - 10px + (60px - 70px))',`

image

PR #648, opened 2025-10-09
Component | Crosshair: Add skipRangeCheck prop

lee00678:
Adding `skipRangeCheck` prop to crosshair.

The new function `onCrosshairMove` will always check if mouse is within range. This will cause certain tests to not recognize mouseover events. We are adding this prop for testing purposes.

cc @rokotyan


rokotyan:
Looks good!

Issue #607, opened 2025-07-21
[Vue/Nuxt] SSR & reactivity breaking


ineshbose:
Hopefully this link works:
https://stackblitz.com/edit/github-riylyusg?file=app%2Fapp.vue

If not, attaching repro ZIP: 

[github-riylyusg.zip](https://github.com/user-attachments/files/21352782/github-riylyusg.zip)

Based on the traceback, `@unovis/ts/utils/misc` imports `to-px` which in turn has

```js
var PIXELS_PER_INCH = getSizeBrutal('in', document.body) // 96
```

on the top level, but this won't work in SSR.

---

~~The same reproduction should also highlight issue with reactivity with `data` passed to the map, as there is no watcher on `data` calling `setData` here: https://github.com/f5/unovis/blob/main/packages/vue/src/html-components/leaflet-map/index.vue#L12~~ closed by #630 

---

I also want to point out that `@nuxtjs/color-mode` targets `document` (i.e., `html` rather than `body`, same for `tailwindcss` integrations adding `data-theme="dark"` or `class="dark"` on the top level `html`), so the map styleDark's change observer on `document.body` doesn't integrate too well and a configuration for this would be incredible.

GrahamSH-LLK:
I believe this might be a Nuxt 4 regression- the library worked perfectly fine until I upgraded Nuxt a few minutes ago. Also, looking at to-px, it seems to have a node entrypoint too.

ineshbose:
> I believe this might be a Nuxt 4 regression- the library worked perfectly fine until I upgraded Nuxt a few minutes ago.

Possible. Although based on the traceback and lifecycle, I'm working why Nuxt 3 would work OK; let me investigate on that end.

(Also pointing out the existing reactivity issue, and hopefully `styleDark` configuration)

ineshbose:
> Also, looking at to-px, it seems to have a node entrypoint too.

Ah you're right... and `vite-node` is picking up the browser entry 🤔

dennisadriaans:
@ineshbose Previously, Vite optimized the dependencies correctly, which made to-px run smoothly on SSR. This seems to have been changed in V4. I am trying to learn more here: https://github.com/nuxt/nuxt/issues/32543 and here: https://github.com/nuxt/nuxt/issues/32581

GrahamSH-LLK:
[...]

PR #649, opened 2025-10-10
Component | Leaflet Map: observe documentElement in addition to body for theme changes

ineshbose:
resolves part 3 of #607 

Would this be a breaking change to ship it by default? If so, we can rather provide a config option instead as per maintainers' preference.

### DIFF
--- a/packages/dev/src/examples/maps/leaflet/leaflet-map-css-vars/index.tsx
+++ b/packages/dev/src/examples/maps/leaflet/leaflet-map-css-vars/index.tsx
@@ -19,6 +19,9 @@ export const component = (props: ExampleViewerDurationProps): React.ReactNode =>
 
   return (<>
     <div className={s.map}>
+      <button className={s.toggleButton} onClick={() => document.documentElement.classList.toggle('dark-theme')}>
+        Toggle documentElement
+      </button>
       <VisLeafletMap<MapPointDatum>
         ref={mapRef}
         data={cities}
--- a/packages/dev/src/examples/maps/leaflet/leaflet-map-css-vars/style.module.css
+++ b/packages/dev/src/examples/maps/leaflet/leaflet-map-css-vars/style.module.css
@@ -10,3 +10,10 @@
   --vis-dark-map-point-inner-label-text-color-dark: green;
   --vis-dark-map-point-inner-label-text-color-light: white;
 }
+
+.toggle-button {
+  position: absolute;
+  top: 10px;
+  left: 10px;
+  z-index: 1000;
+}
--- a/packages/shared/examples/custom-nodes-graph/constants.ts
+++ b/packages/shared/examples/custom-nodes-graph/constants.ts
@@ -1,7 +1,7 @@
 import type { Selection } from 'd3-selection'
 import { css } from '@emotion/css'
 import { group, max, mean, min } from 'd3-array'
-import { getNumber, getColor, getString, getTransformValues, Graph, trimString } from '@unovis/ts'
+import { darkThemeCssSelectors, getNumber, getColor, getString, getTransformValues, Graph, trimString } from '@unovis/ts'
 import type { GraphConfigInterface, GraphNode, StringAccessor } from '@unovis/ts'
 import {
   DEFAULT_CIRCLE_LABEL_SIZE,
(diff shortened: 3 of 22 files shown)


ineshbose:

rokotyan:
I like the idea of making the theme-switching configurable, but I think it needs to be done globally for the whole library, allowing to specify the element to observe and the dark theme attribute / class name.

ineshbose:
> I like the idea of making the theme-switching configurable, but I think it needs to be done globally for the whole library, allowing to specify the element to observe and the dark theme attribute / class name.

In that case would you have suggestions on how the library would like to incorporate and track the implementation(s)?

rokotyan:
@ineshbose 
Let's do it the easy way. I would add the following code to `utils/style.ts`:

```ts
export const darkThemeCssSelectors = 'html[data-theme="dark"],html.dark-theme,body.dark-theme,html.theme-dark,body.theme-dark'

export const isDarkThemeEnabled = (): boolean => {
  return document.documentElement.getAttribute('data-theme') === 'dark' ||
    ['dark-theme', 'theme-dark'].some(className =>
      document.documentElement.classList.contains(className) ||
      document.body.classList.contains(className)
    )
}
```

Then you'd want to replace every occurrence of `body.theme-dark` in the CSS code with `${darkThemeCssSelectors}`, and use the `isDarkThemeEnabled` function in LeafletMap.

rokotyan:
You might also want to add a new dev example (`packages/dev/src/examples`) to test this.

khushi-motwani:

ineshbose:
> You might also want to add a new dev example (`packages/dev/src/examples`) to test this.

Hi @rokotyan, @khushi-motwani and I are working on this PR now and adding the example - could you help us in figuring out how to get this integrated and running in docs locally?

More specifically, we are looking to understand:
1. how to have documentElement/body be dark only for this example
2. how to run examples locally in dev (not sure which script to run)

Thank you!

lee00678:
> > You might also want to add a new dev example (`packages/dev/src/examples`) to test this.
> 
> Hi @rokotyan, @khushi-motwani and I are working on this PR now and adding the example - could you help us in figuring out how to get this integrated and running in docs locally?
> 
> More specifically, we are looking to understand:
> 
> 1. how to have documentElement/body be dark only for this example
> 2. how to run examples locally in dev (not sure which script to run)

To run dev locally, if you pulled the latest on main, `pnpm dev` in your root directory. If not, `npm run dev`
To run website, do `pnpm website` or `npm run website`. 
https://github.com/f5/unovis/blob/main/package.json#L21-L22

khushi-motwani:
Thank you for your help, everything is ready for review.

ineshbose:
Could this be reviewed please?

lee00678 (dismissed):
This looks good to me. @rokotyan do you want to take a quick look?

rokotyan:
@ineshbose Thanks for the ping. I ran the code and it works well. I have a few comments what we need to address to merge this:
- Currently we have two dark theme examples for Leaflet Map. I think we should combine them using the "[Light and Dark Theme](http://localhost:9500/examples/Leaflet/Light%20and%20Dark%20Theme)" one as the base
  image
- We want to use `darkThemeCssSelectors` **everywhere** to make the whole library support various ways to set the dark theme. It's a rather small change, but it needs to be done in many places. A good task for an LLM agent. Here's a changelog example how to make it work for Area:
  image
- We'll need to create a clean commit history to follow out commitlint configuration (@lee00678 Maybe you could help wit this)

@ineshbose Will you be able to address the first two points?

lee00678:
> * We'll need to create a clean commit history to follow out commitlint configuration (@lee00678 Maybe you could help wit this)

I'm happy to help with the commit history once this PR has been finalized.

ineshbose:
Thanks for the input - my apologies for the delay. @lee00678 @rokotyan 

I've made the change to my best knowledge, but I fear that it may be redundant seeing recent commits on `main` about migrating to new CSS variables definition syntax. I intend for the scope of my PR is to extend the functionality of the Leaflet Map, so avoiding large diffs or overlapping changes.

Please assist where necessary so we can have this merged in soon; hopefully this week. :-)

lee00678 (dismissed):
Thanks for the contribution! This will be include in our 1.6.6 release.

PR #654, opened 2025-10-15
Component | Crosshair: Update shouldShow check

lee00678:
This is an update for https://github.com/f5/unovis/pull/648/files
Which I think create issue when `skipRangeCheck` is set up `true`, but you also have real mouse `xPx`, your tooltip will always trigger.

https://github.com/user-attachments/assets/11bf3b98-8a32-4b62-9e83-4cc739a85adb


lee00678:
cc @rokotyan

suryahanumandla:
LGTM!

PR #656, opened 2025-10-23
Component | Sankey: Add node collapse feature (1.7)

lee00678:
Adding a feature to Sankey to allow pre-collapsed node. 
Three new props:

```ts
   /** Enable node collapse functionality. When enabled, clicking on nodes will toggle their collapse state. Default: `false` */
  enableNodeCollapse?: boolean;
  /** Node collapse animation duration, ms. Default: `300` */
  collapseAnimationDuration?: number;
  /** Field name in the node data that indicates if a node should be pre-collapsed.
   *  For example, if set to "disabled", nodes with `disabled: true` will start collapsed.
   *  Default: `undefined` */
  disabledField?: string;
```

https://github.com/user-attachments/assets/c1150b1d-c831-4c6e-98ee-6e9adc1879bc

### DIFF
--- a/packages/ts/src/components/sankey/index.ts
+++ b/packages/ts/src/components/sankey/index.ts
@@ -688,46 +688,31 @@ export class Sankey<
   }
 
   /**
-   * Collapses a node by hiding only the links directly connected to it.
-   * All other nodes (including children and descendants) remain visible in their original positions.
-   * Only the immediate incoming and outgoing links of the collapsed node are hidden.
+   * Sets the collapse state of a node and triggers re-rendering.
    */
-  collapseNode (node: SankeyNode<N, L>): void {
+  private _setNodeCollapseState (node: SankeyNode<N, L>, collapsed: boolean): void {
     const { config } = this
 
-    // Clear any active highlights before collapsing
+    // Clear any active highlights before changing state
     this.disableHighlight()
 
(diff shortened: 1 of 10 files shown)


suryahanumandla:
@lee00678  is it expected behavior?


https://github.com/user-attachments/assets/cae23056-0766-41c7-947f-cc06db7795da

packages/ts/src/components/sankey/index.ts
```diff
@@ -353,11 +360,78 @@ export class Sankey<
...
+  expandNode (node: SankeyNode<N, L>): void {
+    const { config } = this
+
+    // Clear any active highlights before expanding
+    this.disableHighlight()
+
+    node._state = node._state || {}
+    node._state.collapsed = false
+    this._render(config.collapseAnimationDuration)
+  }
```
suryahanumandla, line 702: I see common logic in both methods, can we simplify this?

lee00678: Nice catch! Simplified the code.

lee00678:
> @lee00678 is it expected behavior?
> 
>  Screen.Recording.2025-11-06.at.8.25.30.AM.mov

Yes, this is. Because when collapse, we disable `to` and `from` the node.

Issue #661, opened 2025-10-29
Tooltip retains previous content when empty string or undefined is returned


suryahanumandla:
https://github.com/user-attachments/assets/d9dd016b-3ab2-4487-b607-22e8dd5e5346

suryahanumandla:
Closing as fixed.

CC: @lee00678

PR #662, opened 2025-10-29
Component | Tooltip: Fix Tooltip retains previous content when empty string or undefined is returned

suryahanumandla:
ref: https://github.com/f5/unovis/issues/661


lee00678:
https://github.com/user-attachments/assets/d2407e9c-6063-45c3-a307-3059b3f1065b

Verified. This fix seems to be working correctly.

rokotyan:
@suryahanumandla Thanks for fixing this

PR #664, opened 2025-10-31
Component | Graph: Fit View to Bounds

rokotyan:
https://github.com/f5/unovis/pull/657


rokotyan:
I need to pay more attention when creating PRs. This was also supposed to be in another repo. This is the main one https://github.com/f5/unovis/pull/657

PR #665, opened 2025-10-31
Core | Containers: Add onRenderComplete callback

rokotyan:
Introduce a callback in the container configuration to notify when a chart's rendering is complete.

- Add onRenderComplete to ContainerConfigInterface with appropriate types.
- Invoke the callback in SingleContainer and XYContainer render methods after rendering.


rokotyan:
Also wrong repo 🤦‍♂️

PR #676, opened 2025-11-13
Component | Graph: Fix node coordinate flickering

itai-gendler:
## Problem
Nodes flicker to (0,0) or spiral positions during initial render, layout changes, or when adding nodes dynamically. This happens when node coordinates are undefined/NaN.

## Solution
Ensure all nodes have valid x/y coordinates (defaulting to graph center) before layout calculation. This prevents D3's force simulation from initializing nodes at invalid positions.

## Changes
- Added coordinate validation in `_calculateLayout()` before any layout algorithm runs
- Nodes with invalid coordinates are initialized to graph center (`width/2`, `height/2`)

## Testing
- ✅ Initial render
- ✅ Layout switching
- ✅ Dynamic node addition
- ✅ No console errors

### DIFF
--- a/packages/ts/src/components/graph/index.ts
+++ b/packages/ts/src/components/graph/index.ts
@@ -427,6 +427,14 @@ export class Graph<
   private async _calculateLayout (): Promise<void> {
     const { config, datamodel } = this
 
+    // FIX: Always ensure nodes have valid x/y positions before layout calculation
+    // This prevents D3 force simulation from initializing nodes at NaN/undefined
+    // which causes the flickering effect and "undefined coordinates" errors
+    for (const node of datamodel.nodes) {
+      if (!node.x || isNaN(node.x)) node.x = this._width / 2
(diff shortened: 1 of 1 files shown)


itai-gendler:

itai-gendler:
recheck

itai-gendler:
well after more validation. this improves but doesn't fix the issue.
i'm not sure why, but many times when layout is switched there is a flickering occuring.
i am using vue.js.

lee00678:
@itai-gendler thanks for the contribution! We will take a look at this.

rokotyan:
@itai-gendler Do you happen to have a reproducible example? I wonder why the nodes have undefined/NaN coordinates in the first place

lee00678:
> Do you happen to have a reproducible example? I wonder why the nodes have undefined/NaN coordinates in the first place

@itai-gendler could you please provide more info on this? Thanks.

PR #677, opened 2025-11-17
Component | Treemap: Fix update transition

devgru:
`smartTransition` is known to interrupt existing opacity transitions, this change fixes transitions on label groups.


rokotyan:
@devgru Looks like something went wrong with this PR

devgru:
@rokotyan should be fixed now

PR #689, opened 2025-12-17
Component | Axis: Improve axis rendering performances

aguingand:
### Problem 
I've found a page with 9 XY charts to be slow at render. It's seems to be due to multiple calls of `window.getComputedStyle()` when rendering the text of each tick label (this triggers a reflow / recalc).

### Solution
Put the computed style of the first tick label in an object (`_textStyle`). For each future labels, the cached style will be applied. I've also added `fontWidthToHeightRatio` because it was recalculated in in `utils/text` functions multiple times.

### Attachments
Before ~4s:
image

After ~1s:
image

### DIFF
--- a/packages/ts/src/components/axis/index.ts
+++ b/packages/ts/src/components/axis/index.ts
@@ -40,7 +40,7 @@ export class Axis<Datum> extends XYComponentCore<Datum, AxisConfigInterface<Datu
   private _requiredMargin: Spacing
   private _defaultNumTicks = 3
   private _collideTickLabelsAnimFrameId: ReturnType<typeof requestAnimationFrame>
-  private _textStyle: {
+  private _tickTextStyleCached: {
     fontSize: number;
     fontFamily: string;
     fontWidthToHeightRatio: number;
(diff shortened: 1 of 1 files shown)


aguingand:

packages/ts/src/components/axis/index.ts
```diff
@@ -39,6 +40,11 @@ export class Axis<Datum> extends XYComponentCore<Datum, AxisConfigInterface<Datu
   private _requiredMargin: Spacing
   private _defaultNumTicks = 3
   private _collideTickLabelsAnimFrameId: ReturnType<typeof requestAnimationFrame>
+  private _textStyle: {
```
rokotyan:
```suggestion
  private _tickTextStyleCached: {
```

packages/ts/src/components/axis/index.ts
```diff
@@ -282,14 +285,28 @@ export class Axis<Datum> extends XYComponentCore<Datum, AxisConfigInterface<Datu
...
       }
 
-      const textBlock: UnovisText = { text, fontFamily, fontSize }
-      renderTextToSvgTextElement(textElement, textBlock, textOptions)
+      const textBlock: UnovisText = {
+        text,
+        fontFamily: this._textStyle.fontFamily,
+        fontSize: this._textStyle.fontSize,
+        fontWidthToHeightRatio: this._textStyle.fontWidthToHeightRatio,
+      }
```
rokotyan:
```suggestion
       const textBlock: UnovisText = { text, ...this._tickTextStyleCached }
```

packages/ts/src/components/axis/index.ts
aguingand: Done !

packages/ts/src/components/axis/index.ts
aguingand: Done !

lee00678 (approved):
@aguingand thanks for the contribution! Looks good.

PR #693, opened 2025-12-23
Axis: Fix rotated tick text position + more

rokotyan:
This PR includes improvements to text rendering in the Axis component, enhancements to the Annotations component with debugging capabilities, and several developer experience updates including new examples and theme handling.

### Text Utils - dominant-baseline configuration
- Replaced hardcoded `0.8em` `dy` offset in `renderTextToTspanStrings` with configurable `dominant-baseline`
- Fixes text vertical alignment issues where text was incorrectly shifted down

### Annotations Component - Debug Bounding Box
- Added new `renderTextBoundingBoxes` configuration option for rendering text bounding boxes during development
- Improved DOM management by removing annotation subjects when they are not needed
- ⚠️ The newly added example showed that text rotation is not working in annotations (see below)

### Axis - Tick Label Rotation Fix
- Manually calculate tick text `y` position based on rotation angle for more accurate label placement
- Improved text positioning when axis labels are rotated

⚠️ There still a couple issues in the component which we can fix later
1. The axis bounding box calculations are a bit off in preRender, and I couldn't figure out why. This leads to tick labels overlapping with the axis label. It can be manually fixed by increasing the `labelMargin` config property

2. The `_getRequiredMargin` calculations are incorrect when the tick labels are rotated into the chart. It would be nice to fix this, but I don't think it's a significant issue because this configuration is somewhat invalid

#### Before

https://github.com/user-attachments/assets/7e8dedda-3f51-456b-a9b5-08ea8c6e309c

#### After

https://github.com/user-attachments/assets/356f9d69-fb22-4aa3-879d-854c53090be6

### Dev App Theme Selector
- Auto-enable dark theme based on system preferences (`prefers-color-scheme`) and a minor style fix for the input field

### New and Updated Examples
- **Annotation Options**: Comprehensive example showcasing various annotation configuration options
- **Bitcoin Price Peaks**: Real-world example demonstrating annotations on a financial chart (taken from the website gallery)
- **Axis Tick Label Rotations Slider**: New example with interactive slider to test rotation angles from -180° to 180°. 

SCR-20251223-mxqs
SCR-20251223-mxpy
SCR-20251223-mxsd

### DIFF
--- a/packages/ts/src/components/axis/index.ts
+++ b/packages/ts/src/components/axis/index.ts
@@ -246,6 +246,11 @@ export class Axis<Datum> extends XYComponentCore<Datum, AxisConfigInterface<Datu
     selection.selectAll('*').interrupt()
     const transition = smartTransition(selection, duration).call(axisGen)
 
+    // Unset D3's default y and dy attributes because we're going to set them manually in the renderTextToSvgTextElement function
+    selection.selectAll<SVGTextElement, number | Date>('text')
+      .attr('dy', null)
+      .attr('y', null)
+
     // Resolving tick label overlap after the animation is over
     transition.on('end', () => {
       this._resolveTickLabelOverlap(selection)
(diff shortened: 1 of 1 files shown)


lee00678:
@rokotyan I noticed after this update, all the yAxis ticks by default is no longer vertically centered: 
Screenshot 2026-01-12 at 10 41 15 AM

rokotyan:
Totally missed that, will take a look

PR #697, opened 2026-01-06
Component | Sankey: New `onLayoutCalculated` callback and `getSankeyDepth` method; Fixes

rokotyan:
A bunch of small Sankey tweaks and fixes 
- `onLayoutCalculated` config callback
- `getSankeyDepth` public method
-  Constraint pan in `setZoomScale`
-  Better node icon centering; Hide icon when the node is too small (<2px height)
- Fix `getLabelMaxWidth` handling (the switch operator usage was wrong)

### DIFF
--- a/packages/website/docs/networks-and-flows/Sankey.mdx
+++ b/packages/website/docs/networks-and-flows/Sankey.mdx
@@ -267,6 +267,34 @@ which also configures the node's color:
   nodeSort={(n1, n2) => n1.x - n2.x}
   showContext="minimal"/>
 
+## Post-Layout Customization `1.5`
+The `Sankey` component includes an `onLayoutCalculated` callback that runs after the layout is computed. This lets you inspect or adjust nodes and links, or react to the computed layout metrics.
+
+The callback receives:
+- `nodes: SankeyNode[]`
+- `links: SankeyLink[]`
+- `depth: number` — number of layout levels
+- `width: number`, `height: number` — computed layout size
+- `bleed: Spacing` — calculated outer spacing including labels `{ top, right, bottom, left }`
+
+Example:
+```ts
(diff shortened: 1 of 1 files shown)


packages/ts/src/components/sankey/modules/node.ts
```diff
@@ -154,17 +153,20 @@ export function updateNodes<N extends SankeyInputNode, L extends SankeyInputLink
...
-      .attr('text-anchor', 'middle')
-      .style('dominant-baseline', 'central')
-      .style('stroke', (d: SankeyNode<N, L>) => getColor(d, config.nodeIconColor))
-      .style('fill', (d: SankeyNode<N, L>) => getColor(d, config.nodeIconColor))
-      .style('font-size', (d: SankeyNode<N, L>) => {
-        const nodeHeight = d.y1 - d.y0
-        return nodeHeight < s.SANKEY_ICON_SIZE ? `${nodeHeight * 0.65}px` : null
-      })
-      .html((d: SankeyNode<N, L>) => getString(d, config.nodeIcon))
+    nodeIcon.each((d, i, els) => {
```
lee00678, line 156: This is such a small thing but maybe we should call this `nodeIcons`?

lee00678 (approved):
Looks good!

packages/ts/src/components/sankey/modules/node.ts
rokotyan: This seems to be a single icon (i.e. there's `select` instead of `selectAll`) above. But I don't know why we use `.each` after that. Needs some refactoring

lee00678: @rokotyan, going to wait on this fix, before merging.

PR #698, opened 2026-01-06
Frameworks: Using `toString` comparison for functions when comparing props

rokotyan:
Currently, if you fail to memoize your accessor functions in React, Vue, and other frameworks, you'll end up having phantom re-renders.

This PR adds an option to compare functions using their stringified values and integrates it into the `arePropsEqual` functions. Since the accessor functions are generally small and simple, I don't think it'll cause any problems. However, it will bring the significant benefit of not needing to use things like `useMemo` or `useCallback`.

We can also simplify our documentation pages after this.

### DIFF
--- a/packages/solid/src/utils/props.ts
+++ b/packages/solid/src/utils/props.ts
@@ -4,5 +4,5 @@ export const arePropsEqual = <PropTypes>(
   prevProps: PropTypes,
   nextProps: PropTypes
 ): boolean => {
-  return isEqual(prevProps, nextProps)
+  return isEqual(prevProps, nextProps, [], new Set(), true)
 }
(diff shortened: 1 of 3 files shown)


rokotyan:
@zernonia That would be great if you could test it with Vue. I tried with React and it works great.
@hngngn You might be interested in giving it a try with Solid too.

Overall, I'm interested in hearing your opinion regarding this change.

lee00678:
> Overall, I'm interested in hearing your opinion regarding this change.
Do you want to wait till they get back to you before merging this?

rokotyan:
@lee00678 Let's wait a little bit

hngngn:
It seems the update has broken some of the examples in Solid, and possibly Vue and Svelte as well. Currently, I’m only seeing the TopoJSON map example broken—the slider does nothing.

rokotyan:
@hngngn Thanks for checking! Single Container charts definitely got broken, I'll take a look.

rokotyan:
After thinking more about this, I believe it's a bad idea because it can lead to a bunch of rendering issues. For example, when creating an array of accessors using `map`, i.e. `keys.map(key => (d) => d[key])`, the string representation of these functions will look the same and won't capture any updates to the keys if there are any.

It's better to stick to memoization and add a "Performance Hints" section to our documentation.

PR #705, opened 2026-01-20
Website | Bug: Fix website build missing docusurus core issue

lee00678:
It seems @docsearch/core is not installed correctly with `pnpm install`. This PR adds `@docsearch/core` and `@docsearch/react` under `website/package.json`


suryahanumandla:
LGTM!

PR #708, opened 2026-01-21
Component | Stacked Bar: Performance optimizations

rokotyan:
The component was very slow when used with sparse data containing many null values. It was happening for tow reasons:
- We kept creating DOM elements even for 0 height bars
- We were doing data copy operations via the ...spread operator which is slow

This PR fixes these problems and add a new dev example.

### Before
https://github.com/user-attachments/assets/9625b40f-ebeb-4849-9147-862ac4b877b2

### After
https://github.com/user-attachments/assets/8fc999b5-a799-4b8f-ac12-dc32adb21f49


lee00678 (approved):
Looks good!

PR #716, opened 2026-01-27
Component | TopoJSON: Add support for point shapes

lee00678:
This PR adds point shape feature to TopoJSON map.
With default color:
Screenshot 2026-01-27 at 11 30 52 AM

Color override:

https://github.com/user-attachments/assets/c3be3ffd-a3cf-4e0b-9697-69d7459ade19

### DIFF
--- a/packages/ts/src/components/topojson-map/style.ts
+++ b/packages/ts/src/components/topojson-map/style.ts
@@ -1,4 +1,8 @@
 import { css, injectGlobal } from '@emotion/css'
+export const pointPathRing = css`
+  label: point-path-ring;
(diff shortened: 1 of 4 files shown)


packages/ts/src/components/topojson-map/style.ts
suryahanumandla, line 51: add newly added dark theme variables 

```
 --vis-map-point-default-fill-color: var(--vis-dark-map-point-default-fill-color);
 --vis-map-point-default-stroke-color: var(--vis-dark-map-point-default-stroke-color);
 --vis-map-point-ring-fill-color: var(--vis-dark-map-point-ring-fill-color);
```

PR #718, opened 2026-01-28
Misc: Update all npx commands to pnpm equivalent commands and dependency licenses

suryahanumandla:
- Update all npx commands to pnpm equivalent commands
- Update dependency licenses across all workspaces to reflect latest versions
- Resolve cypress tests issue
- Resolve generate command issues in all workspaces

### DIFF
--- a/packages/angular/licences.txt
+++ b/packages/angular/licences.txt
@@ -9,8 +9,18 @@ tslib                              perpetual       0BSD          2.8.1          
 @angular/core                      perpetual       MIT           12.2.17            angular
 @angular/platform-browser          perpetual       MIT           12.2.17            angular
 @angular/platform-browser-dynamic  perpetual       MIT           12.2.17            angular
+@types/d3-brush                    perpetual       MIT           3.0.6              n/a
+@types/d3-drag                     perpetual       MIT           3.0.7              n/a
+@types/d3-geo                      perpetual       MIT           3.1.0              n/a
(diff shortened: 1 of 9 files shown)


packages/react/autogen/tsconfig.json
```diff
@@ -5,8 +5,12 @@
     "strictNullChecks": false,
     "strictPropertyInitialization": false,
     "declaration": false,
+    "baseUrl": "../",
     "types": [
       "node"
     ],
+    "paths": {
+      "@unovis/shared/*": ["../shared/*"]
```
lee00678, line 13: Is this correct? What is this for?

packages/react/tsconfig.json
```diff
@@ -23,6 +23,7 @@
     "paths": {
       "@unovis/react": ["src/"],
       "@unovis/react/*": ["src/*"],
+      "@unovis/shared/integrations/*": ["../shared/integrations/*"],
```
lee00678: What is this for?

packages/react/autogen/tsconfig.json
suryahanumandla: `pnpm generate` command was throwing errors without this configuration

Image

packages/react/tsconfig.json
suryahanumandla: redundant configuration been added for same issue, can be removed.

> `pnpm generate` command was throwing errors without this configuration`

PR #719, opened 2026-01-28
Misc: Reintegrate bumpp

50rayn:
As done before in #489, the [`bumpp`](https://github.com/antfu-collective/bumpp) functionality returned.
Previously, we have some little problem because of npm workspaces. But since we have pnpm integrated 🥳, it will work well now.


50rayn:
Sorry. Have maken a mistake with source branch

PR #720, opened 2026-01-28
Misc: Integrate the bumpp

50rayn:
As done before in #489, the [`bumpp`](https://github.com/antfu-collective/bumpp) functionality returned.
Previously, we have some little problem because of npm workspaces. But since we have pnpm integrated 🥳, it will work well now.


suryahanumandla:
@50rayn thanks for making the changes. I just tried releasing a beta release, it was failing.  

image

- It has updated the versions in package.json seamlessly but we should also update peer dependency version for @unovis/ts in each workspace

50rayn:
@suryahanumandla may be to try `"@unovis/ts": "workspace:^"` for peerDependencies?

lee00678:
@50rayn thanks for the contribution again! We will review this items for 1.7 release.

PR #722, opened 2026-01-29
Component  | TopoJSON: Add flow animation

lee00678:
This PR implements flow animation in TopoJSON map.

- Added several new props to support all features in leaflet-flow map.
- Added new dev example

https://github.com/user-attachments/assets/adebce88-fd37-4e7d-97bf-b2026c2330e3

### DIFF
--- a/packages/dev/src/examples/maps/topojson/flow-map/index.tsx
+++ b/packages/dev/src/examples/maps/topojson/flow-map/index.tsx
@@ -1,0 +1,133 @@
+import React from 'react'
+import { VisSingleContainer, VisTopoJSONMap } from '@unovis/react'
+import { WorldMapTopoJSON } from '@unovis/ts/maps'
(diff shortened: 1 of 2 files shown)


suryahanumandla:
@lee00678 

When zooming in, a grey dot overlaps the point, causing it to appear completely grey instead of its actual color.


https://github.com/user-attachments/assets/b06ce9c7-c9e4-4433-91dc-7b1405803ed3

lee00678:
Should be fixed now


https://github.com/user-attachments/assets/5109a6aa-9577-4ca9-8ad6-c1223ad1095f

suryahanumandla:
I see gap between flow particles for first flow and second flow has little difference, first flow has too close particles comparing to second one.

image

suryahanumandla:
LGTM!

PR #726, opened 2026-02-02
Component | TopoJSONMap : Add support for color map

suryahanumandla:
image

### DIFF
--- a/packages/angular/src/components/topojson-map/topojson-map.component.ts
+++ b/packages/angular/src/components/topojson-map/topojson-map.component.ts
@@ -10,6 +10,8 @@ import {
   ColorAccessor,
   StringAccessor,
   MapPointLabelPosition,
+  TopoJSONMapClusterDatum,
+  TopoJSONMapPointStyles,
 } from '@unovis/ts'
 import { GeoProjection } from 'd3-geo'
 import { VisCoreComponent } from '../../core'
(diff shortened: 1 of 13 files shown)


packages/ts/src/components/topojson-map/index.ts
```diff
@@ -269,11 +274,32 @@ export class TopoJSONMap<
       .style('cursor', d => getString(d, config.pointCursor))
       .style('opacity', 1)
 
+    // Update donut charts and circles
+    const currentZoomLevel = this._currentZoomLevel || 1
+    pointsMerged.select(`.${s.pointDonut}`).each(function (d, i) {
+      const donutData = getDonutData(d, config.colorMap)
+      if (donutData.length > 0) {
+        const radius = getNumber(d, config.pointRadius) / currentZoomLevel
+        const arcWidth = 2 / currentZoomLevel // Keep arc width constant in screen space
```
lee00678: This is not right. If you look at leaflet map, the default color map is a pie chart. With arcWidth fixed, we can't make it a pie chart. 
See here: 
https://github.com/f5/unovis/blob/main/packages/ts/src/components/leaflet-map/modules/node.ts#L90-L93

I think this should follow a similar implementation. this is why I think cluster and color map may need to be implemented together. I think leaflet map consider colormap a `donutCluster`

packages/angular/src/components/topojson-map/topojson-map.component.ts
```diff
@@ -113,6 +115,48 @@ export class VisTopoJSONMapComponent<AreaDatum, PointDatum, LinkDatum> implement
   /** Link target accessor function. Default: `d => d.target` */
   @Input() linkTarget?: ((l: LinkDatum) => number | string | PointDatum)
 
+  /** Flow source point longitude accessor function or value. Default: `f => f.sourceLongitude` */
+  @Input() sourceLongitude?: NumericAccessor<LinkDatum>
```
lee00678, line 119: I wonder why all the flow animation stuff would be here in the angular component

PR #728, opened 2026-02-02
Misc | Release: Fix update version script

lee00678:
This PR fixes two issues:

- [ ] Version update was missing for `/dev` and `/website` since it was not part of the `workspace`
- [ ] `pnpm version ${new_version} --workspaces` this line doesn't understand `workspace:*` and tries to pull the latest(unpublished) version from npm. So instead of using pnpm, we just call perl.

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -57,6 +57,7 @@
     "vue": "^3"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


suryahanumandla:
@lee00678 please do `pnpm install` and add pnpm-lock.yaml file as well, thanks.

PR #732, opened 2026-02-03
Misc: Vulnerability fix

lee00678:


package.json
```diff
@@ -76,7 +76,26 @@
...
+      "webpack@>=5.0.0-alpha.0 <5.94.0": ">=5.94.0",
+      "http-proxy-middleware@<2.0.7": ">=2.0.7",
+      "semver@>=7.0.0 <7.5.2": ">=7.5.2",
+      "esbuild@<=0.24.2": ">=0.25.0",
+      "@babel/runtime@<7.26.10": ">=7.26.10",
+      "webpack-dev-server@<=5.2.0": ">=5.2.1",
+      "tmp@<=0.2.3": ">=0.2.4",
+      "postcss@<8.4.31": ">=8.4.31",
+      "node-forge@<1.3.2": ">=1.3.2",
+      "@angular/common@<19.2.16": ">=19.2.16",
```
lee00678: I don't think we can upgrade this right? angular common? @suryahanumandla

suryahanumandla: Not required.

suryahanumandla: ideally, we should fix all vulnerabilities with respect to workspace rather at here at global level.

PR #733, opened 2026-02-03
Component | TopoJSON: Add area label to TopoJSON map

lee00678:
- Add `areaLabel` to topoJSON config. This enables user to pass label to areas, it comes with collide detection similar to scatter plot
- Add dev example with areaLabels

https://github.com/user-attachments/assets/e9cc329a-9877-4494-abb6-beeac0327dc1

### DIFF
--- a/packages/ts/src/components/topojson-map/config.ts
+++ b/packages/ts/src/components/topojson-map/config.ts
@@ -3,10 +3,9 @@ import { ComponentConfigInterface, ComponentDefaultConfig } from 'core/component
 
 // Types
 import { ColorAccessor, NumericAccessor, StringAccessor } from 'types/accessor'
-import { GenericDataRecord } from 'types/data'
 
 // Local Types
-import { MapPointLabelPosition, TopoJSONMapPointShape, TopoJSONMapPointStyles } from './types'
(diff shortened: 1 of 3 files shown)


suryahanumandla:
- Probably not an issue, when zoom in and zoom-out take a look at 11th second in the attached screen cast . You can see lot of labels being collisioned and removed after 1 or 2 seconds later. It would be great if we could get rid off it.


https://github.com/user-attachments/assets/52baa230-b3aa-4e4e-8955-fcdc65e8647e

- Any reasons for update pnpm-lock.yml file?

lee00678:
> * Probably not an issue, when zoom in and zoom-out take a look at 11th second in the attached screen cast . You can see lot of labels being collisioned and removed after 1 or 2 seconds later. It would be great if we could get rid off it.
> 
This is because we don't calculate collideAreaLabels during zoom, so if you zoom out quickly, it will wait till zooming is done then calculate the labels.

>  Screen.Recording.2026-02-05.at.8.03.02.PM.mov 
> * Any reasons for update pnpm-lock.yml file?

I think I did a install at some point due to some errors, probably not needed.

PR #737, opened 2026-02-10
Misc | Vue: Fix autogen script and build issue

suryahanumandla:
Fix vue build issue

```
src/components/crosshair/index.vue.ts:12:1 - error TS4082: Default export of the module has or is using private name 'Props'.

 12 export default (<Datum,>(
    ~~~~~~~~~~~~~~~~~~~~~~~~~
 13 __VLS_props: Awaited<typeof __VLS_setup>['props'],
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
... 
116 })(),
    ~~~~~
117 ) => ({} as import('vue').VNode & { __ctx?: Awaited<typeof __VLS_setup> }))type __VLS_NonUndefinedable<T> = T extends undefined ? never : T;
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/components/timeline/index.vue.ts:13:1 - error TS4082: Default export of the module has or is using private name 'Props'.

 13 export default (<Datum,>(
    ~~~~~~~~~~~~~~~~~~~~~~~~~
 14 __VLS_props: Awaited<typeof __VLS_setup>['props'],
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
... 
117 })(),
    ~~~~~
118 ) => ({} as import('vue').VNode & { __ctx?: Awaited<typeof __VLS_setup> }))type __VLS_NonUndefinedable<T> = T extends undefined ? never : T;
    ~~~
```

### DIFF
--- a/packages/vue/autogen/component.ts
+++ b/packages/vue/autogen/component.ts
@@ -33,8 +33,9 @@ import { arePropsEqual, useForwardProps } from '../../utils/props'
 ${isStandAlone ? '' : `import { ${elementSuffix}AccessorKey } from '../../utils/context'\n`}
 ${isStandAlone ? '' : `const accessor = inject(${elementSuffix}AccessorKey)\n`}
 // data and required props ${isComplexPropComponent ? '\n// !!! temporary solution to ignore complex type. related issue: https://github.com/vuejs/core/issues/8412' : ''}
-${isComplexPropComponent ? `interface Props extends /** @vue-ignore */ ${componentName}ConfigInterface${genericsStr} { }` : `type Props = ${componentName}ConfigInterface${genericsStr}`}
-const props = defineProps<Props & { data?: ${dataType} }>()
+${isComplexPropComponent
+    ? `const props = defineProps</** @vue-ignore */ ${componentName}ConfigInterface${genericsStr} & { data?: ${dataType} }>()`
+    : `type Props = ${componentName}ConfigInterface${genericsStr}\nconst props = defineProps<Props & { data?: ${dataType} }>()`}
 
 ${propDefs.length && !isStandAlone ? `${propDefs.join('\n')}` : isStandAlone ? 'const data = computed(() => props.data)' : ''}
 // config
--- a/packages/shared/integrations/utils.ts
+++ b/packages/shared/integrations/utils.ts
@@ -75,6 +75,13 @@ export function getTypeName (type: ts.Node | undefined): string {
       const generics = (type as ts.TypeReferenceNode).typeArguments?.map(getTypeName).join(', ')
       return name + (generics ? `<${generics}>` : '')
     }
+    case (ts.SyntaxKind.TemplateLiteralType): {
+      const t = type as ts.TemplateLiteralTypeNode
+      const spans = t.templateSpans
+        .map(span => `\${${getTypeName(span.type)}}${span.literal.text}`)
+        .join('')
+      return `\`${t.head.text}${spans}\``
+    }
     case (ts.SyntaxKind.ArrayType): return `${getTypeName((type as ts.ArrayTypeNode).elementType)}[]`
     case (ts.SyntaxKind.TupleType): return `[${(type as ts.TupleTypeNode).elements.map(getTypeName).join(', ')}]`
     default: {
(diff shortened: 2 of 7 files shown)


packages/vue/autogen/component.ts
```diff
@@ -24,6 +24,69 @@ export function getComponentCode (
   // If the build is failing, add the respective component here.
   const complexPropComponent = ['Timeline', 'Crosshair']
   const isComplexPropComponent = complexPropComponent.includes(componentName)
+  const genericsDecl = generics ? `<${generics.map(g => g.extends ? `${g.name} extends ${g.extends}` : g.name).join(', ')}>` : ''
+  const imports = [
+    ...importStatements.map(s => `import { ${s.elements.join(', ')} } from '${s.source}'`),
+    componentName === 'Timeline' ? 'import { XYComponentConfigInterface, StringAccessor } from "@unovis/ts"' : '',
```
lee00678: Why are we hardcode this for `Timeline`?

rokotyan: Yeah, look wrong, I don't think other integrations have that. In my opinion any overrides like this should be avoided. We can modify the component itself if needed to make it work.

suryahanumandla: @lee00678 @rokotyan I thought we intentionally kept these components so added the check, thanks for the suggestion let me make it more generic.

https://github.com/f5/unovis/blob/c1944ce742a1906ca449c25ab41670791ef95dd4/packages/vue/autogen/component.ts#L25

packages/vue/autogen/component.ts
```diff
@@ -20,21 +20,32 @@ export function getComponentCode (
...
     : 'config.value'
 
-  // Vue 3.3.4 has issue resolving complex Typescript, in this case when the type has `WithOptional`.
-  // If the build is failing, add the respective component here.
-  const complexPropComponent = ['Timeline', 'Crosshair']
-  const isComplexPropComponent = complexPropComponent.includes(componentName)
+  // Additional imports needed for specific components
+  const componentSpecificImports: Record<string, string[]> = {
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    Timeline: ['import { XYComponentConfigInterface, StringAccessor } from "@unovis/ts"'],
```
lee00678: I see we did keep `timeline` and `crosshair` separate from the rest because of TS errors. In this update, `crosshair` no longer needs it? Because I only see `timeline` here.

lee00678:
@suryahanumandla I think this update broke something:

https://github.com/user-attachments/assets/03ac5be8-7b72-452d-894e-506f6d16033d

on main:

https://github.com/user-attachments/assets/a5d2375b-dae8-44ed-9c3d-e19e253fc914

packages/vue/autogen/component.ts
suryahanumandla: This is completely different,  I have added an extra functionality where we can pass component specific imports if any. 

Earlier we had a check like this

```typescript
 ${componentName === 'Timeline' ? '\nimport { XYComponentConfigInterface, StringAccessor } from "@unovis/ts"'
```

#### Why crosshair no longer needs it?
>  I see we did keep timeline and crosshair separate from the rest because of TS errors. In this update, crosshair no longer needs it? Because I only see timeline here.

Earlier we had a check whether component has complex types(i.e. isComplexPropComponent), now there is no check required.

```
// data and required props ${isComplexPropComponent ? '\n// !!! temporary solution to ignore complex type. related issue: https://github.com/vuejs/core/issues/8412' : ''}
${isComplexPropComponent ? `interface Props extends /** @vue-ignore */ ${componentName}ConfigInterface${genericsStr} { }` : `type Props = ${componentName}ConfigInterface${genericsStr}`}
```

suryahanumandla:
> @suryahanumandla I think this update broke something:
> 
>  Screen.Recording.2026-03-17.at.1.02.04.PM.mov 
> on main:
> 
>  Screen.Recording.2026-03-17.at.1.00.17.PM.mov

@lee00678 I have added a fix for it. Please review it, thanks.

lee00678 (dismissed):

lee00678 (approved):

PR #747, opened 2026-02-24
Component | Timeline: Label rendering fixes and Docs

rokotyan:
Currently the `rowLabelWidth ` and `rowMaxLabelWidth` properties don't work because of the bug in `trimSVGText`. This PR fixes that along with a few other tweaks:
- Two new config properties: `rowLabelTrimMode` and `rowLabelMargin`
- New dev example to test `rowLabelTrimMode` and `rowLabelMargin`
- Extended clip path to take into account arrowheads when they are in use
- Major update to the docs incorporating all the recent changes

image

image


lee00678 (dismissed):
looks good.

lee00678 (approved):

PR #759, opened 2026-03-30
Component | Axis: Configurable tick size

rokotyan:
Configurable `tickSize` for Axis

```ts
  /** The size of the tick marks in pixels. Accepts a single number (applies to both inner and outer ticks)
   * or a tuple `[innerTickSize, outerTickSize]`. Default: `6` */
  tickSize?: number | [number, number];
```

image


lee00678 (approved):
Looks good

PR #767, opened 2026-04-09
React: Use specific imports instead of barrel 

rokotyan:
Currently, our React components (and other integrations too) import everything from the main entry point `@unovis/ts`. This causes bundlers to load the entire Unovis codebase into memory when compiling a project, which consumes a lot of memory and makes tree-shaking more difficult.

This PR changes the approach to importing only what's needed. This should make the build of projects that use Unovis faster and consume less memory, which is important in CI environments.

I have tried this with one project, and there are no issues. After testing it further with React, we should try extending this approach to other integrations.

### DIFF
--- a/packages/ts/package.json
+++ b/packages/ts/package.json
@@ -30,13 +30,26 @@
   "module": "./index.js",
   "typings": "./index.d.ts",
   "type": "module",
+  "typesVersions": {
+    "*": {
+      "components/*": ["dist/components/*"],
+      "containers/*": ["dist/containers/*"],
+      "core/*": ["dist/core/*"],
+      "types/*": ["dist/types/*"],
+      "utils/*": ["dist/utils/*"],
+      "styles/*": ["dist/styles/*"],
+      "data-models/*": ["dist/data-models/*"],
+      "data/*": ["dist/data/*"]
+    }
+  },
   "sideEffects": [
     "styles/index.js",
     "dist/styles/index.js"
   ],
   "scripts": {
-    "build": "sha=$(tar cf - ./src | shasum); if [[ $(echo $sha) == $(< .srcsha) ]] && [[ -d \"./dist\" ]]; then echo \"Lib Build Exists\"; else npm run forcebuild; echo $sha > .srcsha; fi",
-    "forcebuild": "rimraf dist; rollup -c; rm -rf dist/.cache; cp LICENSE README.md package.json ./dist",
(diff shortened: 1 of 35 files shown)


lee00678:
@rokotyan do you have any suggestions in terms of how to verify, testing this? The best way I can think of is to create a react application and run it there with a beta release?

packages/ts/package.json
```diff
@@ -30,13 +30,26 @@
...
   "sideEffects": [
     "styles/index.js",
     "dist/styles/index.js"
   ],
   "scripts": {
-    "build": "sha=$(tar cf - ./src | shasum); if [[ $(echo $sha) == $(< .srcsha) ]] && [[ -d \"./dist\" ]]; then echo \"Lib Build Exists\"; else npm run forcebuild; echo $sha > .srcsha; fi",
-    "forcebuild": "rimraf dist; rollup -c; rm -rf dist/.cache; cp LICENSE README.md package.json ./dist",
+    "build": "sha=$(tar cf - ./src | shasum); if [[ $(echo $sha) == $(< .srcsha) ]] && [[ -d \"./dist\" ]]; then echo \"Lib Build Exists\"; else npm run forcebuild; echo $sha > .srcsha; fi; npm run build:package-json",
+    "forcebuild": "rimraf dist; rollup -c; rm -rf dist/.cache; cp LICENSE README.md ./dist; npm run build:package-json",
+    "build:package-json": "node -e \"const p=JSON.parse(require('fs').readFileSync('package.json','utf8')); delete p.typesVersions; require('fs').writeFileSync('./dist/package.json', JSON.stringify(p, null, 2))\"",
```
suryahanumandla, line 53: @rokotyan why we need this?

suryahanumandla (commented):
LGTM!

> verified after rebasing with main, all respective workspace commands are working perfect.

packages/ts/package.json
rokotyan: @suryahanumandla I needed to add that to make types work in our dev setup, but I didn't want these fields to be shipped with the dist. I couldn't find a more elegant solution and open to suggestions of you have any

```json
 "typesVersions": {
    "*": {
      "components/*": ["dist/components/*"],
      "containers/*": ["dist/containers/*"],
      "core/*": ["dist/core/*"],
      "types/*": ["dist/types/*"],
      "utils/*": ["dist/utils/*"],
      "styles/*": ["dist/styles/*"],
      "data-models/*": ["dist/data-models/*"],
      "data/*": ["dist/data/*"]
    }
  },
  ```

rokotyan:
> @rokotyan do you have any suggestions in terms of how to verify, testing this? The best way I can think of is to create a react application and run it there with a beta release?

@lee00678 I verified it in a large React app where I use Unovis extensively.

suryahanumandla:
@rokotyan please resolve merge conflicts, MR is good for merge. thanks

lee00678 (approved):

PR #769, opened 2026-04-20
Component | Topojson: Fix pan and zoom then click node to resize position bug

lee00678:
This PR fix the issue when zoom and pan the map, then click on a node, and resize the chart, the node sometimes goes off the screen.


suryahanumandla (approved):
LGTM!

Issue #633, opened 2025-09-09
Is there a way to force an XYContainer to redraw?


davestewart:
I've built a l little Unovis chart debugger, with UI inputs that plug into XYContainer, StackedBar and Axis, props.

However, it seems that some attributes (such as `tickPadding`) don't seem to update dynamically, and I have to remove and re-add the chart to see changes.

Is this by design, am I missing something, or is there a way to tell the chart / components to redraw after an attribute changes?

I'm using Vue-flavoured Unovis.

davestewart:
Also weirdly earlier, I have a set of debugging attributes I add to the various components:

```ts
{
  bar: {
    [VisGroupedBarSelectors.root]: {
      class: 'unovis-grouped-bar',
    },
    [VisStackedBarSelectors.root]: {
      class: 'unovis-stacked-bar',
    },
  },
  axis: {
    [VisAxisSelectors.root]: {
      class: 'unovis-axis',
    },
    [VisAxisSelectors.tickLabel]: {
      class: 'unovis-tickLabel',
    },
  },
}
```

If I add them in the markup, then I can still set something like `tickValues` and the chart updates:

```vue
<VisXYContainer
  :data="data"
  v-bind="chartOptions.container"
>
  <VisStackedBar
    v-bind="chartOptions.bar"
    :attributes="{
      [VisStackedBarSelectors.root]: {
        class: 'unovis-stacked-bar',
      },
    }"
  />
  <VisAxis
    type="x"
    v-bind="chartOptions.axis.x"
    :attributes="{
      [VisAxisSelectors.root]: {
        class: 'unovis-axis',
      },
      [VisAxisSelectors.tickLabel]: {
        class: 'unovis-tickLabel',
      },
    }"
  />
```

But if I add them in code (i.e. a big object that I `v-bind` to the individual components, then if I update `tickValues` then the chart _doesn't_ update, until I re-render it.

I'm investigating whether this is something to do with Vue's reactivity system, or something else.

If anyone has any debugging tips to share, such as "here's how to see what the chart is receiving and what functions should fire when props update" then I would be glad to hear it.

davestewart:
I added `tickPadding` as a reactive prop directly, and it does not seem to update when changed.

davestewart:
Current workaround is to re-render the chart; temporarily setting the animation `duration` to 0:

```ts
function redraw() {
  const duration = model.value.chart.container.duration
  model.value.chart.container.duration = 0
  model.value.debug.renderId++
[...]

PR #773, opened 2026-04-28
Vue | Crosshair: Fix `forceShowAt` not redrawing

50rayn:
While building a synced two-panel chart in `@unovis/vue`, I noticed `forceShowAt` updates never redraw the crosshair until the panel itself gets a pointer event. React, Svelte, Solid all work — only Vue is broken.

Two things going on:

- `useForwardProps` reads `getCurrentInstance().attrs`, which is the raw (non-reactive) object. Spreading it in the computed never tracks reactivity, so the child's `watch(config, ...)` never fired on prop change. Switched to `useAttrs()`.
- Even with the watcher firing, `Crosshair.setConfig(...)` doesn't redraw on its own. Added `component.value?.render()` after `setConfig` (Crosshair-only, in autogen).

Added `shared/examples/synced-crosshairs/` (Vue + Svelte + Solid) with a toolbar for poking `forceShowAt` programmatically — easier to repro than hovering. React example already exists.

Note: the `useForwardProps` fix also unblocks runtime prop changes for all other wrappers (`VisLine`, `VisArea`, …) — silent bug, just less noticeable.


packages/vue/autogen/component.ts
```diff
@@ -60,7 +60,7 @@ onUnmounted(() => {
 
 watch(config, (curr, prev) => {
   if (!arePropsEqual(curr, prev)) {
-    component.value?.setConfig(config.value)
+    component.value?.setConfig(config.value)${componentName === 'Crosshair' ? '\n    component.value?.render()' : ''}
```
rokotyan: I wonder why the render is only for Crosshair? I think other components will need it too, unless it's being called by the Container, like we do in our React wrapper.

rokotyan (commented):
This is great, thanks @50rayn

packages/vue/autogen/component.ts
50rayn: Good catch — every wrapper hits the same path: child `setConfig` mutates a raw class instance, so the parent `XYContainer` watchEffect never sees a dependency change and `updateContainer` doesn't fire. Dropped the Crosshair-only condition in autogen and regenerated all wrappers.

lee00678: I recall there was a conversation about hardcode `crosshair` and `timeline` when we tried to fix the vue autogen. https://github.com/f5/unovis/pull/737#discussion_r2796365499

lee00678:
@50rayn this is great! The gallery example is not complete, could you please finish that? [https://unovis.dev/contributing/guides/gallery-examples](https://unovis.dev/contributing/guides/gallery-examples) steps are here. I think would be great if we could have a demo on the website. 

I've merged #776 and rebased this branch, please take a look and make sure everything looks right.

50rayn:
> @50rayn this is great! The gallery example is not complete, could you please finish that? https://unovis.dev/contributing/guides/gallery-examples steps are here. I think would be great if we could have a demo on the website.
> 
> I've merged #776 and rebased this branch, please take a look and make sure everything looks right.

1. Rebased
2. Added the example _Synced Crosshairs_ example in the gallery.

rokotyan:
@lee00678 Once we merge this, can we release a beta? I think this fix will also address https://github.com/f5/unovis/issues/633#issuecomment-4105889728

Also I think we should add automatic beta releases from CI after every push to `main`. This way the community won't need to wait till we publish a proper version.

50rayn:
> @lee00678 Once we merge this, can we release a beta? I think this fix will also address [#633 (comment)](https://github.com/f5/unovis/issues/633#issuecomment-4105889728)
> 
> Also I think we should add automatic beta releases from CI after every push to `main`. This way the community won't need to wait till we publish a proper version.



> @lee00678 Once we merge this, can we release a beta? I think this fix will also address [#633 (comment)](https://github.com/f5/unovis/issues/633#issuecomment-4105889728)
> 
> Also I think we should add automatic beta releases from CI after every push to `main`. This way the community won't need to wait till we publish a proper version.

i do something like:

* https://github.com/Noction/vue-use-flexsearch/blob/main/.github/workflows/release.yml
* https://github.com/Noction/vue-use-flexsearch/blob/main/.github/workflows/ci.yml

lee00678:
> @lee00678 Once we merge this, can we release a beta? I think this fix will also address [#633 (comment)](https://github.com/f5/unovis/issues/633#issuecomment-4105889728)
> 
> Also I think we should add automatic beta releases from CI after every push to `main`. This way the community won't need to wait till we publish a proper version.

Yes. Released [1.6.6-beta.0](https://www.npmjs.com/package/@unovis/vue/v/1.6.6-beta.0)

lee00678 (approved):
Great!

PR #774, opened 2026-04-28
Misc: Drop `.npmrc` legacy-peer-deps flag

50rayn:
Splitting another small piece off #495.

The `.npmrc` had `legacy-peer-deps=true`, which is an npm-only flag — and we're on pnpm, so it was a no-op anyway. @rokotyan also said in the comments that he didn't remember why it was there and to try removing it.

Verified `pnpm install` still works clean. Since the file only contained that one line, dropped the file altogether.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,0 @@
-legacy-peer-deps=true


lee00678:
I think this makes sense, I don't recall us using `npm` for anything anymore. cc @suryahanumandla could you please check and merge? Thanks.

suryahanumandla (approved):
LGTM! we no longer using npmrc

PR #775, opened 2026-04-28
Misc: Migrate husky to v9

50rayn:
Husky 7 was printing a deprecation warning on every commit telling us to drop the shebang lines from `.husky/pre-commit` and `.husky/commit-msg` or it'll fail in v10.

So I bumped husky to v9, dropped the shebang/sourcing lines (v9 just runs the file content directly), and updated `postinstall.sh` to use `pnpm exec husky` instead of `pnpm dlx husky install` (the `install` subcommand is deprecated and `dlx` re-downloads it on each install).

Verified that pre-commit (lint-staged) and commit-msg (commitlint) still run on this commit.

### DIFF
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,4 +1,1 @@
-#!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
-
-pnpm exec commitlint --edit ${1}
+pnpm exec commitlint --edit $1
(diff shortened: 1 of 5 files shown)


lee00678 (approved):
Tested by checking out a new repo, `pnpm i` and committing some bad message.
Also tested by cutting a new branch from this feature branch, both case works.

Thanks for update!

PR #776, opened 2026-04-28
Vue | Wrappers: Drop unused `parseProps`

50rayn:
Spotted while working on #773 — `parseProps` in `packages/vue/src/utils/props.ts` is exported but not used anywhere across the repo (`grep -rn 'parseProps\b' packages/vue` confirms it's dead). TypeScript was complaining (`TS6133: 'props' is declared but its value is never read`).

Removed the function and the now-orphan `ComponentInternalInstance` import.


lee00678 (approved):
Looks good. Thanks for the fix!

PR #778, opened 2026-05-01
Website: Deploy dev examples

rokotyan:
This PR adds a Code panel to our dev examples and deploys them under the `/examples` sub-path. We might want to clean them up a little bit before announcing, e.g. not all of the LeafletMap examples are working.

The dev code is only for React but it still can be useful for the community.

Live demo: https://rokotyan.github.io/unovis/examples/

https://github.com/user-attachments/assets/d7da0b12-bd21-4af9-938b-826701847fa0


lee00678 (dismissed):
This is great! I think this will be super helpful to the community.

Yes, agree with doing some clean up.

PR #782, opened 2026-05-03
Container: Add `load` / `render` / `redraw` lifecycle events

50rayn:
Wanted a reliable way to know when a chart actually finished drawing — for logging first-paint times and reacting to renders downstream. The only existing hook was the `onRenderComplete` function-prop callback, which feels foreign in Vue/Svelte/Angular where events are the idiom. The usual workaround is `await nextTick()` + a couple of `requestAnimationFrame`s, but that's a guess, not a real signal from unovis.

So I wired up `load` (first render), `render` (every render), `redraw` (every render except first) on `VisXYContainer` and `VisSingleContainer` — Highcharts-style. Vue gets `@load`, Svelte `on:load`, Angular `(load)`. React/Solid keep `onRenderComplete` (function-prop callbacks are idiomatic there).

While at it, refactored the underlying `XYContainer.onRenderComplete` / `SingleContainer.onRenderComplete` from 7 positional args to a single payload object — that way `XYContainerRenderPayload` / `SingleContainerRenderPayload` live once in `@unovis/ts` and everyone imports them. This is a small breaking change for anyone passing `onRenderComplete` directly to the TS class or to React/Solid wrappers.

Verified `pnpm check` for Vue/Svelte, `ng build` for Angular, and the TS package rebuilds clean.

### DIFF
--- a/packages/ts/src/containers/single-container/config.ts
+++ b/packages/ts/src/containers/single-container/config.ts
@@ -7,6 +7,15 @@ import { Annotations } from '@/components/annotations'
 // Types
 import { Spacing } from '@/types/spacing'
 
+export type SingleContainerRenderPayload = {
+  svg: SVGSVGElement;
+  margin: Spacing;
+  containerWidth: number;
+  containerHeight: number;
+  componentWidth: number;
+  componentHeight: number;
+}
+
 export interface SingleContainerConfigInterface<Datum> extends ContainerConfigInterface {
   /** Visualization component. Default: `undefined` */
   component?: ComponentCore<Datum>;
--- a/packages/website/docs/containers/Single_Container.mdx
+++ b/packages/website/docs/containers/Single_Container.mdx
@@ -210,5 +210,43 @@ _ChordDiagram_ all accept a `pattern` accessor, and the `theme-patterns` class a
 automatically. See [Patterns](/docs/guides/theming#patterns) in the _Theming_ guide.
 :::
 
+## Events
+The Vue, Svelte, and Angular wrappers emit lifecycle events whenever the chart renders.
+The payload is the same `SingleContainerRenderPayload` for all three:
+
+```ts
+type SingleContainerRenderPayload = {
+  svg: SVGSVGElement;
+  margin: Spacing;
+  containerWidth: number;
+  containerHeight: number;
+  componentWidth: number;
+  componentHeight: number;
+}
+```
+
+| Event    | When it fires                                       |
+| -------- | --------------------------------------------------- |
+| `load`   | Once, after the chart's first render completes.     |
+| `render` | After every render, including the first.            |
+| `redraw` | After every render except the first.                |
+
+```vue
+<!-- Vue -->
+<VisSingleContainer :data @load="onLoad" @render="onRender" @redraw="onRedraw" />
+```
+
+```svelte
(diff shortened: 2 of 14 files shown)


packages/ts/src/containers/single-container/config.ts
```diff
@@ -15,14 +24,7 @@ export interface SingleContainerConfigInterface<Datum> extends ContainerConfigIn
...
   /** Callback function to be called when the chart rendering is complete. Default: `undefined` */
-  onRenderComplete?: (
-    svgNode: SVGSVGElement,
-    margin: Spacing,
-    containerWidth: number,
-    containerHeight: number,
-    componentWidth: number,
-    componentHeight: number,
-  ) => void;
+  onRenderComplete?: (payload: SingleContainerRenderPayload) => void;
```
rokotyan: Switching to a payload object will be a breaking change, so we won't be able to do it till we release version 2.0. We'll need to stick to arguments

packages/website/docs/containers/Single_Container.mdx
```diff
@@ -130,5 +130,43 @@ Sankey</a> supports the `sizing` option. By default, all components will fit to
...
+  containerWidth: number;
+  containerHeight: number;
+  componentWidth: number;
+  componentHeight: number;
+}
+```
+
+| Event    | When it fires                                       |
+| -------- | --------------------------------------------------- |
+| `load`   | Once, after the chart's first render completes.     |
```
rokotyan, line 230: Do you think we really need three of these events instead of just one? I think the users can track whether it was the first render themselves if they need to. Could you give a couple of use cases to justify that?

rokotyan (commented):
I like the idea! But there're a couple things I'm concerned about.

packages/website/docs/containers/Single_Container.mdx
50rayn: I was inspired by Highcharts (`chart.events.load` / `render` / `redraw`).  Since I'm currently migrating from it, I have a couple of moments where I use these. Use cases I had in mind:

- `load` — the big one. Fires once, perfect for prefetching chart settings, hydrating a "chart settings" panel, logging first-paint time,  kicking off a fade-in, and hiding a skeleton. A userland boolean works, but every team ends up rewriting it and getting it subtly wrong (re-mounts, keyed components, hot reload, props that retrigger updates without a true "first paint").
- `redraw` — react only to data/prop updates without the noisy first paint. e.g. block a "chart settings" toggle button while a redraw is in flight, refit an overlay, restore scroll, re-sync an external legend.
- `render` — superset, blocks UI on every paint (first + redraw). Useful when you want a generic "chart is busy" gate around the whole lifecycle without branching on first vs subsequent.
 
Cost on our side is one boolean per container, so emitting all three is essentially free. Happy to drop redraw if it really feels redundant — but I'd push hard to keep `load` since it's the one consumers actually trip over. WDYT?

packages/ts/src/containers/single-container/config.ts
50rayn: Reverted the TS-core `onRenderComplete` to positional args; wrappers build `XYContainerRenderPayload` / `SingleContainerRenderPayload` internally before emitting. No breaking change for existing _TS_ / _React_ / _Solid_ users.

PR #788, opened 2026-05-07
Component | TopoJSON: Map fit to point function need to consider source and target node

lee00678:
This is a fix for map fit to point in TopoJSON
When there's flow data, the source and target node should be included as well

https://github.com/user-attachments/assets/e4ad4f5c-d8db-4f98-81aa-b7db32cbfdbd


suryahanumandla:
@lee00678 can you please update dev example?, thanks

lee00678:
> @lee00678 can you please update dev example?, thanks

I didn't since we are planning on making the dev examples available for 1.6.6. I think the one without the flow looks better. I would just add a target/source node and test if possible.

suryahanumandla (approved):
LGTM!

PR #798, opened 2026-05-09
Dev: Replace `seedrandom` with `d3-random`

50rayn:
Matches #495's swap. `seedrandom` (unmaintained since 2018) → `@faker-js/faker` for the sample data generator in `dev/utils/data.ts`. `randomNumberGenerator()` keeps the same call signature (returns a number in `[0, 1]`); seeded via `faker.seed(123)` so example data stays deterministic.

Sample-data values shift to a new (still deterministic) sequence — galleries will look slightly different but no example breaks. Webpack dev build verified working.

### DIFF
--- a/packages/dev/licences.md
+++ b/packages/dev/licences.md
@@ -5,11 +5,11 @@
 | react                                | perpetual      | MIT          | 19.2.4            | n/a                                                                |
 | react-dom                            | perpetual      | MIT          | 19.2.4            | n/a                                                                |
 | react-router-dom                     | perpetual      | MIT          | 6.30.3            | Remix Software <hello@remix.run>                                   |
-| seedrandom                           | perpetual      | MIT          | 3.0.5             | David Bau                                                          |
 | serve                                | perpetual      | MIT          | 14.2.6            | n/a                                                                |
 | @emotion/css                         | perpetual      | MIT          | 11.13.5           | Kye Hohenberger                                                    |
 | d3-array                             | perpetual      | ISC          | 3.2.4             | Mike Bostock http://bost.ocks.org/mike                             |
 | d3-format                            | perpetual      | ISC          | 3.1.2             | Mike Bostock http://bost.ocks.org/mike                             |
+| d3-random                            | perpetual      | ISC          | 3.0.1             | Mike Bostock http://bost.ocks.org/mike                             |
 | @percy/cli                           | perpetual      | MIT          | 1.31.9            | n/a                                                                |
 | @percy/cypress                       | perpetual      | MIT          | 3.1.7             | Perceptual Inc.                                                    |
 | @pmmmwh/react-refresh-webpack-plugin | perpetual      | MIT          | 0.5.17            | Michael Mok                                                        |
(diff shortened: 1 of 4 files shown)


rokotyan:
@50rayn Did you have any problem with `seedrandom`? It's a pretty narrow use-case library, so I see no issues with it being updated infrequently. I remember that faker was a pretty heavy library and it might be an overkill to use it for such a simple problem. Another alternative — [random](https://github.com/transitive-bullshit/random#readme)

50rayn:
> @50rayn Did you have any problem with `seedrandom`? It's a pretty narrow use-case library, so I see no issues with it being updated infrequently. I remember that faker was a pretty heavy library and it might be an overkill to use it for such a simple problem. Another alternative — [random](https://github.com/transitive-bullshit/random#readme)

`seedrandom` is CJS — needed an `import * as` + `new` shim in _data.ts_. _@faker-js/faker_ is ESM, so the import is clean.

It also opens up realistic generators (names, currencies, dates) we can extend the dev examples with, and lean on in the unit/integration tests we want to add. devDep only.

`randomNumberGenerator()` keeps the same `() => [0, 1]` signature, seeded via `faker.seed(123)`. Sample data shifts to a new deterministic sequence — galleries look slightly different, nothing breaks.

rokotyan:
@50rayn How's about using `d3-random` https://d3js.org/d3-random#randomLcg since we already depend heavily on D3?

50rayn:
> @50rayn How's about using `d3-random` https://d3js.org/d3-random#randomLcg since we already depend heavily on D3?

Good catch, @rokotyan! My initial idea was to also use `fakerjs` for generating names, countries, and other sample data across the examples, which would have justified the dependency. Ended up only needing the PRNG, so `d3-random` is clearly the right fit here. Updated the PR.

rokotyan:
@50rayn Nice, thanks for updating!

lee00678 (dismissed):

lee00678:
@50rayn do you mind checking this? Cursor said it was due to a token mismatch. But this has never happened on any of your other PRs. 

```Root cause: fork PR
I found a matching failed run on branch dev/replace-seedrandom-with-faker ([run 26045954795](https://github.com/f5/unovis/actions/runs/26045954795)). That branch is [PR #798](https://github.com/f5/unovis/pull/798) from fork 50rayn/unovis (isCrossRepository: true).

For pull requests from forks, GitHub intentionally gives GITHUB_TOKEN read-only access, even if the workflow requests write permissions. CodeQL cannot upload SARIF results in that case.```

50rayn:
> @50rayn do you mind checking this? Cursor said it was due to a token mismatch. But this has never happened on any of your other PRs.
> 
> ```
> I found a matching failed run on branch dev/replace-seedrandom-with-faker ([run 26045954795](https://github.com/f5/unovis/actions/runs/26045954795)). That branch is [PR #798](https://github.com/f5/unovis/pull/798) from fork 50rayn/unovis (isCrossRepository: true).
> 
> For pull requests from forks, GitHub intentionally gives GITHUB_TOKEN read-only access, even if the workflow requests write permissions. CodeQL cannot upload SARIF results in that case.```
> ```

No problem. Checking 🫡.

50rayn:
Very strange. I rebased and updated the license files.

lee00678 (approved):

PR #804, opened 2026-05-12
Misc: Bump rollup-plugin-typescript2 to ^0.37.0

50rayn:
Bumps `rollup-plugin-typescript2` in `@unovis/ts` from `^0.31.1` to `^0.37.0`. Dep tree shrinks net (drops `@yarn-tool/resolve-package`, `ts-toolbelt`, `resolve`; adds only `semver`).

`0.37` is stricter about declaration emit — the inline `as unknown as TopoJSON.Topology<{...}>` casts in `src/maps.ts` trigger TS2742 because the inferred type's reference to `@types/topojson-specification` goes through pnpm's `.pnpm/` directory and isn't portable. Refactored to a single generic `Topo<K extends string>` alias + `topo<T>(json)` helper; type-level only, `dist/maps.js` is byte-identical.

`react`/`angular`/`svelte`/`vue` also pin `0.31` but stay there for now — bumping them surfaces a separate TS6059 workspace-symlink `rootDir` issue that needs its own look.

### DIFF
--- a/packages/ts/licences.md
+++ b/packages/ts/licences.md
@@ -74,7 +74,7 @@
 | rollup-plugin-postcss             | perpetual      | MIT          | 4.0.2             | EGOIST <0x142857@gmail.com>                                       |
 | rollup-plugin-rename-node-modules | perpetual      | MIT          | 1.3.1             | Lazyuki                                                           |
 | rollup-plugin-terser              | perpetual      | MIT          | 7.0.2             | Bogdan Chadkin <trysound@yandex.ru>                               |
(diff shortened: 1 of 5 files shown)


lee00678:
I think we can close this in favor of https://github.com/f5/unovis/pull/869?

50rayn:
> I think we can close this in favor of #869?

yep. what was a small step to Vite. since it's already done - it can be closed.

lee00678:
Close in favor of https://github.com/f5/unovis/pull/869

PR #805, opened 2026-05-13
Component | TopoJSON: Add flow source into mapFitToPoints calculation

lee00678:
This PR: 
- Fixes the mapfitToPoint issue, previously it was not factor in the source node position
- Updating the dev example with flow and make it a little better

https://github.com/user-attachments/assets/259ff470-52ff-483e-808a-0f41425ea725


suryahanumandla (approved):
LGTM!

PR #806, opened 2026-05-15
Misc | Build: Publish framework packages in place

50rayn:
## What

- Switch `@unovis/{ts,react,vue,svelte,solid}` from transplant publish (copy into `dist/`, publish from there) to in-place publish (publish from package root), matching how `@unovis/angular` already works
- Entry points are now `dist/`-prefixed (`"main": "./dist/index.js"` etc.), `"files": ["dist"]` whitelists tarball contents — npm still auto-includes `package.json`, `README`, and `LICENSE` from root
- Drop the `cp … ./dist` build step and `publish:dist` scripts; root `publish:*` now call `pnpm publish` directly
- `@unovis/svelte` also strips the extra `package.json`/`LICENSE`/`README` that `@sveltejs/package` emits into `dist/`
- Mark `@unovis/dev` as `private` (it's the examples app, never meant to be published)

## Why

The framework packages' `.d.ts` files contain `typeof import("@unovis/ts/dist/components/.../style")` references — 25 of them in `@unovis/react@1.6.5` alone. But the released `@unovis/ts` ships everything at package root with no `dist/` folder, so those paths are dead in the currently published packages. This publish convention change fixes that as a side effect: `@unovis/ts` now ships a `dist/` folder, so the imports resolve correctly.

Verified with `npm pack` diff against `1.6.5` — react/vue/svelte/solid file lists match exactly (just relocated under `dist/`); type checks clean; consumer resolution test (tarballs extracted into `node_modules`, entry points + `exports` subpaths resolved) passed.


50rayn:
Closing in favor of #767, which landed after this was opened and addresses the same area with a different approach — keeping `@unovis/ts` published from `dist/` and adding `typesVersions` + `build:package-json` to support subpath imports.

This PR's in-place approach (`files: ["dist"]`) produces a nested published layout (`@unovis/ts/dist/components/...`), which is incompatible with the subpath imports #767 introduced. `publishConfig.directory` was also explored, but it changes pnpm's workspace linking to consume `dist/` instead of source — conflicting with the source-based dev setup. The existing `cd dist` publish flow is the right trade-off given those constraints.

PR #813, opened 2026-05-22
Misc | Perf: Text measurement caching 

rokotyan:
### Why
`getComputedTextLength()`, `getComputedStyle().fontSize`, and the off‑DOM SVG measurement used by `getPreciseStringLengthPx` all force synchronous layout / style recalc. Charts with hundreds of labels (Treemap, Sankey, TopoJSON, Scatter) pay this cost per label on every render, dominating frame time.

### What changed
- **`packages/ts/src/utils/text.ts`** — added a small caching layer and switched to canvas measurement:
  - `measureTextWidth(str, font)` *(new export)* — measures via a shared, never-mounted `<canvas>` 2D context. No DOM, no reflow. **~100× faster** than `getComputedTextLength` in a micro-benchmark (5 ms vs 540 ms for 1000 calls).
  - `getCachedFontSizePx(node)` *(new export)* — replaces `parseFloat(getComputedStyle(el).fontSize)` callers. Uses the class-chain cache so `getComputedStyle` is called once per distinct class chain, not once per node (~900 → ~5 calls for the 800-tile stress demo).
  - `clearTextMeasurementCache()` *(new export)* — for web-font load / restyle edge cases.
  - Internal `getCachedComputedTextLength` wraps the four direct `getComputedTextLength` callsites in `wrapSVGText` / `trimSVGText` / `estimateTextSize` with a two-tier cache (per-node WeakMap + global FIFO Map) backed by canvas measurement.
  - `getPreciseStringLengthPx` now uses canvas first, off-DOM SVG only as SSR fallback.
  - All `Map`-backed caches are bounded (FIFO, 5000) to avoid leaks in long-running dashboards; `WeakMap` caches self-collect.

- **Component callsite updates** to use the new helper:
  - `components/treemap/index.ts`
  - `components/scatter/modules/utils.ts` (2 callsites)
  - `components/topojson-map/utils.ts`

- **`components/sankey/modules/label.ts`** — flipped `fastEstimatesMode` to `false` so Sankey now uses the precise (now cached) path.

- **New dev demo**: `packages/dev/src/examples/misc/treemap/many-labels-stress/` — 800-tile tall treemap for profiling the wrap/trim hot path.

### Behavior
Visual output is unchanged in all manually-verified examples (Treemap, Sankey, Scatter, TopoJSON). Canvas widths can differ from SVG `getComputedTextLength` by sub-pixel amounts, well inside the existing 1.1 tolerance in `trimSVGText`.

### Before
image

### After (2x faster rendering)
SCR-20260522-kurd

### DIFF
--- a/packages/dev/src/examples/misc/treemap/many-labels-stress/index.tsx
+++ b/packages/dev/src/examples/misc/treemap/many-labels-stress/index.tsx
@@ -1,0 +1,65 @@
+import React from 'react'
+import { VisSingleContainer, VisTreemap } from '@unovis/react'
+import { FitMode } from '@unovis/ts'
+
+export const title = 'Treemap: Many Labels Stress Test'
+export const subTitle = 'Text-measurement cache stress test'
+
+type TreemapExampleDatum = {
+  name: string;
+  value: number;
+  group?: string;
+  subgroup?: string;
+}
+
+const WORDS = [
+  'Alpha', 'Bravo', 'Charlie', 'Delta', 'Echo', 'Foxtrot', 'Golf', 'Hotel',
+  'India', 'Juliet', 'Kilo', 'Lima', 'Mike', 'November', 'Oscar', 'Papa',
+  'Quebec', 'Romeo', 'Sierra', 'Tango', 'Uniform', 'Victor', 'Whiskey',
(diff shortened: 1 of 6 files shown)


rokotyan:
Oops, too early for the upstream. Will close this for now

PR #817, opened 2026-05-27
Core: Text rendering performance and precision improvements (1.7)

rokotyan:
Our charts measure text widths constantly (for wrapping, trimming, sizing labels). We were re-measuring the same text over and over via slow browser layout, and using an "every character is the same width" shortcut that was imprecise and trimmed labels in the wrong spots sometimes.

This PR addresses those issues:
- Measure text with a lightweight canvas tool and cache the results (capped in size, auto-cleared when web fonts load).
- Default to real measurements instead of the equal-width guess, so labels truncate at the right place.
- Build SVG text elements directly instead of serializing to strings and re-parsing them → faster UnovisText rendering.

### Before
SCR-20260522-kwkf

### After
SCR-20260522-kurd


devgru (approved):
Looks good

lee00678 (approved):

PR #822, opened 2026-06-03
Shared: Gallery Dev Server to preview multiple frameworks

rokotyan:
A gallery dev server that allows you to preview how an example works with different frameworks.

Angular is missing for now because it's "special". We'll need to run a dedicated dev server for it and embed via and iframe.

SCR-20260603-nrkk

### DIFF
--- a/.claude/launch.json
+++ b/.claude/launch.json
@@ -1,0 +1,18 @@
+{
+  "version": "0.0.1",
+  "configurations": [
+    {
+      "name": "dev",
+      "runtimeExecutable": "npm",
+      "runtimeArgs": ["run", "dev"],
+      "cwd": "packages/dev",
+      "port": 9500
+    },
+    {
+      "name": "dev:gallery",
+      "runtimeExecutable": "pnpm",
+      "runtimeArgs": ["run", "dev:gallery"],
+      "port": 9600
+    }
+  ]
+}
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
   },
   "scripts": {
     "dev": "cd packages/dev && pnpm dev",
+    "dev:gallery": "pnpm --filter @unovis/shared dev",
     "website": "cd packages/website && pnpm start",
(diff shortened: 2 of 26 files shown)


lee00678:
Oh don't we love iframe. 😺

rokotyan:
## Build failure root cause + fix 🔍

The `build` check was failing in the **website step** (`docusaurus build --no-minify`) with a Node heap OOM (`exit code 134`). Here's what was actually going on.

### Why the first attempts didn't stick
- **Heap bump** (`NODE_OPTIONS=--max-old-space-size=8192`) *did* take effect — CI then died at **8 GB** instead of 4 GB, just ~20 min later. The build leaks memory **without bound**, so raising the ceiling only delays the crash.
- **Pinning React to 18** wasn't the cause either.

### Actual root cause (confirmed by local bisection)
The website source and all example bundling are byte-identical to `main` — the only thing this PR changes that the Docusaurus build consumes is `packages/shared`'s **dependency graph**. Adding the new deps back one group at a time:

| `packages/shared` deps | website build |
| --- | --- |
| `main` baseline | ✅ builds (~27s) |
| `+` React runtime (`react` / `react-dom` / `@types/react*`) | ✅ builds (~4s) |
| `+` framework **`peerDependencies`** (`@unovis/react` / `vue` / `solid` / `svelte`) | ❌ **OOM** |

Declaring `@unovis/react/vue/solid/svelte` as `peerDependencies` on `@unovis/shared` creates `packages/shared/node_modules/@unovis/*` symlinks that derail webpack's module resolution when it bundles the shared example files, sending it into unbounded memory growth.

These peer deps aren't needed: the gallery dev server already resolves each framework package via explicit `resolve.alias` entries in `vite.config.ts`.

### Fix
Reverted `peerDependencies` to just `@unovis/ts` (+ regenerated lockfile). **Locally verified**: the website now builds in ~4s, comfortably under a 3 GB heap cap.

> Note: the earlier React `^18` pin and the Node heap bump are still on the branch — both harmless, neither is required by this fix.

lee00678 (approved):
Amazing addition!

PR #829, opened 2026-06-15
Color Synchronization & Patterns (1.7)

rokotyan:
Adds two related theming capabilities to Unovis:

1. **Color synchronization** — a chart-wide color function plus per-series *color keys*, so the same data category keeps the same color across multiple charts, legends, and tooltips regardless of how the series are ordered in each chart.
2. **Patterns** — stripe/dot/hatch fills and line markers as a second visual channel on top of color, useful for color-blind-safe palettes, grayscale printing, and dense charts. Patterns can be set explicitly per series/datum, or applied automatically via a theme class.

### Color synchronization
- A container-level `color` function — `(key: string | number) => string` — on both `XYContainer` and `SingleContainer`, applied to every component inside.
- A `colorKeys: string[]` property on components, aligned with the `y` accessors, that labels each series with a stable key.
- `UnovisColorScale` — the default D3 ordinal color scale (backed by the `--vis-color*` CSS variables), exported so it can be re-ranged globally.
- `BulletLegend` now accepts a `color` function and a `colorKey` per item, so legend bullets match the charts.
- Color resolution order per element: explicit `color` accessor → `colorKey` via the color function → series index via the color function.

https://github.com/user-attachments/assets/bba01476-5746-49b9-b49b-4d541cc67c54

### Patterns
- A `pattern` accessor on **Area, Line, Scatter, GroupedBar, StackedBar, Timeline, Donut, NestedDonut, Sankey, and ChordDiagram**.
- `FillPatternType` (`StripesDiagonal`, `Dots`, `StripesVertical`, `Crosshatch`, `Waves`, `Circles`) and `LinePatternType` (`Circle`, `Triangle`, `Diamond`, `Arrow`, `Square`, `Star`) enums for the built-in patterns. Built-in pattern defs are injected once on load.
- Fill patterns apply as an SVG `mask` (inheriting the series color); line patterns combine a `marker` with a `stroke-dasharray`. Markers inherit the line color via `context-stroke`, with a per-SVG fallback for browsers that don't support it (Safari/iOS).
- Automatic mode: adding the `theme-patterns` class to `document.body` applies a default pattern palette by series index. An explicit `pattern` accessor always overrides the automatic styles.
- The automatic palette is customizable through `--vis-pattern-fill{i}`, `--vis-pattern-marker{i}`, and `--vis-pattern-dasharray{i}` CSS variables.
- `BulletLegend` bullets render the matching pattern.

image

## Usage

Synchronizing colors across charts:
```tsx
import { Scale } from '@unovis/ts'

const color = Scale.scaleOrdinal().range(['#4d8cfd', '#ff6b7e', '#00c19a'])
const keys = ['aws', 'azure', 'github']

<VisBulletLegend items={keys.map(name => ({ name, colorKey: name }))} color={color} />

<VisXYContainer data={data} color={color}>
  <VisGroupedBar x={d => d.x} y={keys.map(k => d => d[k])} colorKeys={keys} />
</VisXYContainer>
```

Applying patterns per series:
```tsx
import { FillPatternType } from '@unovis/ts'

const fillPatterns = [FillPatternType.StripesDiagonal, FillPatternType.Dots, FillPatternType.Crosshatch]

<VisStackedBar x={d => d.x} y={[y0, y1, y2]} pattern={(_d, i) => fillPatterns[i]} />
```

## Backwards compatibility
All changes are additive. The `ColorAccessor` type gained an optional `key` third argument (non-breaking), and new container/component properties default to `undefined`. Existing charts are unaffected.

## Documentation
- **Theming guide**: new *Synchronizing Colors Across Charts* section and an expanded *Patterns* section (the `pattern` accessor + the `theme-patterns` class).
- **Container docs**: `color`, `tooltip`, `crosshair`, `annotations`, `duration` (animation), `autoMargin`, and `ariaLabel` are now documented with live examples; Single Container gets a component-level color section.
- **Gallery examples**: *Color Management* (synchronized charts, custom palette) and *Patterns* (fill & line patterns across chart types).

https://github.com/user-attachments/assets/beaad421-cabd-474a-814c-e75b93ade317


devgru (approved):

lee00678 (dismissed):
This is great!

lee00678 (approved):

PR #834, opened 2026-06-19
Component | Area: Fix gradient clipping with non-zero y-domain (#548)

### LINKED ISSUE #548: VisArea gradient doesn't work properly with non-zero yDomain
When using a custom `yDomain` that doesn't start at zero, the gradient in `VisArea` component doesn't render properly. The background gradient seems to be calculated from `y=0` rather than from the minimum value in the yDomain.

### Minimal reproduction (Vue):

```vue
<template>
  <VisXYContainer :data="data" :y-domain="[145, undefined]">
    <VisLine :x="d => d.x" :y="d => d.y" />
    <VisArea :x="d => d.x" :y="d => d.y" color="url(#gradient)" />
  </VisXYContainer>
</template>

<script setup>
const svgDefs = `
  <linearGradient id="gradient" gradientTransform="rotate(90)">
    <stop offset="0%" stop-color="#3366FF" stop-opacity="1" />
    <stop offset="100%" stop-color="#3366FF" stop-opacity="0" />
  </linearGradient>
`;

// Sample data with y values that don't include zero
const data = [
  { x: 'Jan', y: 150 },
  { x: 'Feb', y: 145 },
  { x: 'Mar', y: 155 },
  { x: 'Apr', y: 148 }
];
</script>
```

codeplott:
Fixes #548 

When a custom `y-domain` doesn't include `0`, `VisArea` is still drawn down to the `0` baseline, far outside the visible range. The fix clamps the area to the y scale's range. Default zero-based case is unaffected.

Before:
image

After:
image

### DIFF
--- a/packages/ts/src/components/area/index.ts
+++ b/packages/ts/src/components/area/index.ts
@@ -7,7 +7,7 @@ import { interpolatePath } from 'd3-interpolate-path'
 import { XYComponentCore } from 'core/xy-component'
 
 // Utils
-import { getNumber, getString, isArray, isNumber, getStackedExtent, getStackedData, filterDataByRange, getValue } from 'utils/data'
+import { getNumber, getString, isArray, isNumber, getStackedExtent, getStackedData, filterDataByRange, getValue, clamp } from 'utils/data'
 import { smartTransition } from 'utils/d3'
 import { getColor } from 'utils/color'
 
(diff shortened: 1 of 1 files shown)


codeplott:

codeplott:
recheck

lee00678 (approved):
Looks good! cc @rokotyan

PR #847, opened 2026-07-01
Misc: Percy test update

lee00678:
Added an error message, to see if percy actually run. Because we have been getting a green in CI/CD but no screenshots uploaded. Not getting error message in cypress either.


suryahanumandla (approved):
LGTM!

PR #849, opened 2026-07-03
Dev | Examples: Add framework logos to gallery panel headers

50rayn:
Adds the framework logos (React / Vue / Solid / Svelte / TypeScript) from [Iconify's `logos` set](https://icones.js.org/) to the dev gallery's per-framework panel headers, so it's easier to scan which panel is which.

- SVGs saved under `packages/shared/gallery-dev-server/assets/`
- wired via `` in `index.html`

Dev-tooling only — no library or published-package changes.

2026-07-03 at 12 11 28

PR #850, opened 2026-07-03
Dev | Gallery: Add Angular panel to multi-framework playground

50rayn:
Adds the 6th gallery panel — Angular — so every example renders side-by-side across React, Vue, Solid, Svelte, vanilla TS, and Angular.

Angular 12 predates AnalogJS, so no AOT — it compiles in-browser via View Engine JIT (`platformBrowserDynamic` + `@angular/compiler`):

- `unovis-angular-jit` Vite plugin (`enforce: 'pre'`): inlines `templateUrl`/`styleUrls`, normalizes self-closed custom tags (VE JIT rejects them), and transpiles decorators with `ts.transpileModule` + `emitDecoratorMetadata` — esbuild can't emit it, and it's needed for constructor DI (`DomSanitizer`, `ChangeDetectorRef`) via `reflect-metadata`.
- `mounts/angular.ts`: loads `zone.js` + `@angular/compiler` dynamically only when an Angular panel mounts (other panels untouched); bootstraps a host component whose template holds the example's selector.
- `@unovis/angular` aliased to `src/public-api.ts`, so it resolves without a prior `ng build`.

Dev-server only — no published package changes. All example types render (XY, single-container, graph, maps), incl. the 2 constructor-DI examples.

> The framework-logos commit is from #849; this PR is stacked on top of it.


rokotyan:
This is awesome!

lee00678 (approved):
This is great!

PR #852, opened 2026-07-14
Misc | Webiste, Vue, cmds-report updates

lee00678:
This PR fixes the following issues: (See commits for each issue)

- The Vue components were importing types from incorrect paths:
	import type { GraphInputLink, GraphInputNode } from '@/types/graph'
        import type { GenericDataRecord } from '@/types/data'
   Updating [packages/shared/integrations/utils.ts](https://github.com/f5/unovis/compare/qian/fix-path-issue?expand=1#diff-2b33b0a4e1d1c771657adbdcc4b758bbde73ffa450cea790dfe287042f0da9ad) fixes this issue

- Website build was failing, updated docusaurus config file for the fix.
- We have a few anchor that's broken, this PR also addresses that.
- cmds report Phase 4 auto add dev:gallery and the build will hang, removed that from the script.

### DIFF
--- a/packages/angular/src/components/boxplot/boxplot.component.ts
+++ b/packages/angular/src/components/boxplot/boxplot.component.ts
@@ -80,6 +80,11 @@ export class VisBoxplotComponent<Datum> implements BoxplotConfigInterface<Datum>
   /** Box fill color accessor function. Default: `d => d.color` */
   @Input() color?: ColorAccessor<Datum>
 
+  /** Array of data color keys. Use to map data keys to colors.
+   * Expected to the same length as the `y` accessors array.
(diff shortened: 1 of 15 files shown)


packages/angular/licences.md
```diff
@@ -1,3 +1,5 @@
+../..                                    |  WARN  Unsupported engine: wanted: {"node":">=24.0.0"} (current: {"node":"v20.19.0","pnpm":"10.23.0"})
```
suryahanumandla: Can we get rid off this message in all licences, looks like node pointed to 20 instead of 24.

suryahanumandla (approved):
LGTM!

PR #854, opened 2026-07-20
Misc: Fix dev gallery issue

suryahanumandla:
# pnpm dev:gallery  command got broken After TypeScript 5.5 Upgrade

## Problem

The Vite-based dev gallery (multi-framework preview server) broke after upgrading to TypeScript 5.5+. The root cause: `svelte-preprocess@4` hardcodes the `importsNotUsedAsValues` compiler option, which was removed in TS 5.5, causing a compile error on startup.

## Impact

| Area | Before | After |
|------|--------|-------|
| Dev gallery startup | ❌ Crashes (TS 5.5 error) | ✅ Starts correctly |
| `@/types/` aliases | ❌ Not resolved (wrong prefix) | ✅ Resolved |
| Angular Treemap | ❌ Not importable from `@unovis/angular` | ✅ Exported |
| Angular RadialBar | ❌ Not importable from `@unovis/angular` | ✅ Exported |
| `LeafletMapClusterDatum` | ❌ Missing from public API | ✅ Exported |
| Angular dual-axis-chart | ❌ Second axis rendered outside chart | ✅ Correctly overlaid |
| Vue basic-boxplot | ❌ Empty chart (no data) | ✅ Renders correctly |

## Before
image

## After
image

### DIFF
--- a/packages/ts/src/containers/xy-container/index.ts
+++ b/packages/ts/src/containers/xy-container/index.ts
@@ -433,12 +433,14 @@ export class XYContainer<Datum> extends ContainerCore {
 
   private _getMargin (): Spacing {
     const { config: { margin } } = this
+    const defaultMargin = { top: 0, bottom: 0, left: 0, right: 0 }
+    const m = margin || defaultMargin
 
     return {
-      top: margin.top + this._axisMargin.top,
-      bottom: margin.bottom + this._axisMargin.bottom,
-      left: margin.left + this._axisMargin.left,
-      right: margin.right + this._axisMargin.right,
+      top: m.top + this._axisMargin.top,
+      bottom: m.bottom + this._axisMargin.bottom,
+      left: m.left + this._axisMargin.left,
+      right: m.right + this._axisMargin.right,
     }
   }
 
--- a/packages/ts/src/containers/xy-container/index.ts
+++ b/packages/ts/src/containers/xy-container/index.ts
@@ -437,10 +437,10 @@ export class XYContainer<Datum> extends ContainerCore {
     const m = margin || defaultMargin
 
     return {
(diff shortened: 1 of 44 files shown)


lee00678:
I did another round of checks, fixed the following:
Sometimes margin is undefined, so we can't just call _getMargin()
Screenshot 2026-07-24 at 9 59 38 AM

Instead of '65h' it should be '65vh'
Screenshot 2026-07-24 at 10 01 50 AM

The svelte demo for the dev:gallery page is not loading, due to was wrapped in `<template>`, fixed that
Screenshot 2026-07-24 at 10 25 45 AM

lee00678 (dismissed):

packages/ts/src/containers/xy-container/index.ts
```diff
@@ -433,12 +433,14 @@ export class XYContainer<Datum> extends ContainerCore {
...
     const { config: { margin } } = this
+    const defaultMargin = { top: 0, bottom: 0, left: 0, right: 0 }
+    const m = margin || defaultMargin
 
     return {
-      top: margin.top + this._axisMargin.top,
-      bottom: margin.bottom + this._axisMargin.bottom,
-      left: margin.left + this._axisMargin.left,
-      right: margin.right + this._axisMargin.right,
+      top: m.top + this._axisMargin.top,
```
rokotyan: I feel there might be a potential bug here, since we define Spacing properties as optional:
```ts
export type Spacing = {
  left?: number;
  right?: number;
  top?: number;
  bottom?: number;
}
```

I would do `top: (m.top || 0) + this._axisMargin.top`, etc ...

lee00678: Updated

PR #856, opened 2026-07-22
Misc: Fix breaking changes from the color function API

rokotyan:
This PR addresses two breaking changes introduced by the recent custom color function work, before they ship in a release.

### Rename `color` to `colorFunction`

The new custom color function config property on containers and `BulletLegend` was named `color`, which is both ambiguous (it holds a function, not a color value) and collides with the widely used `color` accessor naming on components. It is now `colorFunction`:

```ts
// Before
<VisXYContainer color={myColorScale}>
// After
<VisXYContainer colorFunction={myColorScale}>
```

Applied to `ContainerConfigInterface` (so both `SingleContainer` and `XYContainer`), `BulletLegendConfigInterface`, and the Angular wrappers — which also now actually expose the input on `VisSingleContainerComponent` and `VisXYContainerComponent`. Dev examples and website docs (`Single_Container`, `XY_Container`, theming guide) are updated to match.

### Backward compatibility for `getColor`

The `getColor` utility's fourth positional argument used to be `dontFallbackToCssVar: boolean` and was repurposed as `key: string`, silently breaking existing callers. `getColor` now accepts `string | boolean` there: a string is treated as a color key, while `true` is interpreted as the legacy `dontFallbackToCssVar` flag (unless explicitly set via `options`). Marked with TODOs to drop the boolean form in Unovis 2.0.

### DIFF
--- a/packages/angular/src/containers/single-container/single-container.component.ts
+++ b/packages/angular/src/containers/single-container/single-container.component.ts
@@ -1,7 +1,7 @@
 import { Component, ViewChild, ElementRef, AfterViewInit, Input, OnDestroy, SimpleChanges, ContentChild } from '@angular/core'
 
 // Vis
-import { ComponentCore, SingleContainer, SingleContainerConfigInterface, Tooltip, Spacing, Annotations, Sizing } from '@unovis/ts'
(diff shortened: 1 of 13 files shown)


lee00678:
I will review on friday and make sure this get merged before 1.7

rokotyan:
@lee00678 I'll also test it locally and then mark as "ready for review"

rokotyan:
Tested, worked well

lee00678 (approved):
LGTM

Issue #855, opened 2026-07-22
VisCrosshair template called with undefined when using data property


crystalfp:
In an unspecified update the VisCrosshair "template" property does not access anymore its "data" property. It works only if I put the data property on the VisXYContainer. I cannot do this because there are two data sets in my chart and because it has always worked this way.
The code below shows a tooltip with "undefined" in it.
```
const template = (d: DataRecord): string => `<div>${JSON.stringify(d)}</div>`;

  <VisXYContainer :margin="{right: 20, top: 30, left: 20, bottom: 10}"
                  :xDomain="range" :yDomain="[0, 100]" :duration="0"
                  :scaleByDomain="true" class="layout-main chart-container">
    <VisLine :key="forceUpdate" :data="points" :x="xp" :y="yp" :curveType/>
    <VisScatter :data="scatter" :x="xp" :y="yp"
                color="red" :size="4"
                :label="lp" labelColor="red" labelPosition="right" />
    <VisAxis type="x" label="2θ (degrees)" :gridLine="showGrid"
              labelColor="#6C778C" :labelFontSize="24" numTicks="10"
              tickTextColor="#6C778C"/>
    <VisAxis type="y" label="Intensity (a. u.)" :gridLine="showGrid"
              labelColor="#6C778C" :labelFontSize="24"
              tickTextColor="#6C778C"/>
    <VisTooltip />
    <VisCrosshair :data="points" :x="xp" :y="yp" :template />
  </VisXYContainer>

```
    "@unovis/ts": "^1.6.7",
    "@unovis/vue": "^1.6.7",
Thanks for looking!

crystalfp:
Also, if I put `:data="points"` on the VisXYContainer, The VisScatter ignores its own `:data="scatter"` property and takes the one on the container.
So seems to me that has changed the priority of the "data" properties: from own then container to contained overrides the own property.

crystalfp:
Returning to 1.6.5 show error also in the accessor for x: `const xp = (d: DataRecord): number => d.x;` when hovering the chart. BDW, the VisLine that access the same dataset works correctly. Continuing to go back in the versions...

crystalfp:
With 1.6.4 the crosshair worked perfectly. So something has changed between 1.6.4 and 1.6.5

rokotyan:
This PR should address the issue 
- #857

crystalfp:
1.7.0 solved the problem. Thanks!

PR #857, opened 2026-07-22
Vue | Crosshair | Timeline | Boxplot: Fix data prop dropped by misplaced @vue-ignore annotation

rokotyan:
Addresses #855

Since 1.6.5, the Vue wrappers for the three "complex prop" components (Crosshair, Timeline, Boxplot) compiled with no runtime props at all, so their own `data` prop was silently ignored: `props.data` was always undefined and, for example, the crosshair tooltip template received `undefined` unless data was set on the container instead.

Root cause: commit d4a22d21 changed the autogen output from an interface-extends pattern to an inline intersection with the ignore annotation in front:
```
  defineProps</** @vue-ignore */ Config<Datum> & { data?: Datum[] }>()
```

Babel attaches a leading comment at the start of a type to the whole TSIntersectionType node, and @vue/compiler-sfc checks for @vue-ignore on a node before recursing into its members - so the entire type, including { data?: Datum[] }, resolved to zero props.

Fix: reorder the intersection so the annotation attaches only to the config member:
```
  defineProps<{ data?: Datum[] } & /** @vue-ignore */ Config<Datum>>()
```
The compiled components now declare `props: { data: {} }` again, byte-identical to the 1.6.4 output, while config props keep flowing through as fallthrough attrs (which useForwardProps reads from attrs). The emitted d.ts still exposes the full Config<Datum> & { data } typing.

Note: reverting to the 1.6.4 interface-extends form is not possible - current vue-tsc fails it with TS4082 (private name 'Props' in the default export), which is the build error that motivated d4a22d21. The reordered intersection avoids both problems.

### DIFF
--- a/packages/vue/autogen/component.ts
+++ b/packages/vue/autogen/component.ts
@@ -34,7 +34,7 @@ ${isStandAlone ? '' : `import { ${elementSuffix}AccessorKey } from '../../utils/
 ${isStandAlone ? '' : `const accessor = inject(${elementSuffix}AccessorKey)\n`}
 // data and required props ${isComplexPropComponent ? '\n// !!! temporary solution to ignore complex type. related issue: https://github.com/vuejs/core/issues/8412' : ''}
 ${isComplexPropComponent
-    ? `const props = defineProps</** @vue-ignore */ ${componentName}ConfigInterface${genericsStr} & { data?: ${dataType} }>()`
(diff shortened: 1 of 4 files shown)


rokotyan:
Maybe we can release 1.6.8 with this fix @lee00678

lee00678 (approved):
Looks good. All build passes.

PR #858, opened 2026-07-24
Component | Donut: Fix jump before animation when switching between full and half

Vincentdevreede:
## Summary
Animate the group, label, and background transforms together with the arc segments instead of setting them instantly, so switching angleRange (e.g. full to half) morphs continuously instead of snapping to the new layout before the arcs animate. Also fixes the background's `d` attribute being interpolated as a plain string, which produced invalid arc flags mid-transition (`Expected arc flag ('0' or '1')` console errors), by tweening the underlying angle/radius state and rebuilding the path each tick like the arc segments already do via `updateArc`.

## Preview
ezgif-57d33d47bb15fd15

### DIFF
--- a/packages/ts/src/components/donut/index.ts
+++ b/packages/ts/src/components/donut/index.ts
@@ -1,4 +1,6 @@
 import { Selection } from 'd3-selection'
+import { Transition } from 'd3-transition'
+import { interpolate } from 'd3-interpolate'
 import { pie, arc } from 'd3-shape'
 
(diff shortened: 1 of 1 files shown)


Vincentdevreede:

rokotyan:
@Vincentdevreede Nice tweak, thanks for the contribution!

lee00678 (approved):
Nice fix! Thanks for the contribution

PR #859, opened 2026-07-24
Component | Tooltip: Fix fade-in animation and follow-cursor freeze over trigger gaps

Vincentdevreede:
## Problems

  **1. Fade-in didn't always animate.** `_display()` removes the `hidden` class and adds the `show` class in the same synchronous tick. Without a paint in between, the browser can coalesce both changes into one style recalc, so the opacity transition has no "before" state to animate from. So the tooltip just appears at full opacity instead of fading in.

  **2. Tooltip froze while fading out over gaps.** With `followCursor: true`, moving the cursor from a trigger into a gap between triggers (e.g. between bars, or space covered by an axis/another sibling component) doesn't fire any of the per-trigger `mousemove` listeners, since each one only listens on its own component's element. The tooltip would freeze at its last position instead of tracking the cursor while it fades out.

## Fixes

**For (1):** insert `this.element.getBoundingClientRect()` between the two `classed()` calls in `_display()` to force a reflow, guaranteeing the browser paints the pre-transition frame before `show` is applied.

**For (2):** `Tooltip._setUpEvents()` already sets up per-component delegated `mousemove`/`mouseleave` listeners. This extends that same loop to also attach to the shared container, so gap movements are seen too. When such a move doesn't match any configured trigger and the tooltip is still fading out, we reposition it to keep it tracking the cursor.

The `hidden` class (not `_isShown`) gates the reposition: `_isShown` flips to `false` the instant `_hide()` starts the fade, while `hidden` is only added once the fade-out has actually finished (in the `transitionend` callback). Gating on `_isShown` would silently skip repositioning for the whole visible fade duration.

No new event listener lifecycle is introduced. This reuses the existing delegation mechanism, adding one more attachment target and one fallback branch.

## Preview

ezgif-8df5a684c6acc341

### DIFF
--- a/packages/ts/src/components/tooltip/index.ts
+++ b/packages/ts/src/components/tooltip/index.ts
@@ -135,9 +135,9 @@ export class Tooltip {
 
   private _display (): void {
     window.clearTimeout(this._hideDelayTimeoutId)
-    this.div
-      .classed(s.hidden, false) // The `hidden` class sets `display: none;`
-      .classed(s.show, true) // The `show` class triggers the opacity transition
+    this.div.classed(s.hidden, false) // The `hidden` class sets `display: none;`
+    this.element.getBoundingClientRect() // force a reflow so it has a frame to fade in from
+    this.div.classed(s.show, true) // The `show` class triggers the opacity transition
 
     this._isShown = true
   }
@@ -348,8 +348,11 @@ export class Tooltip {
     // We use the Event Delegation pattern to set up Tooltip events
     // Every component will have single `mousemove` and `mouseleave` event listener functions, where we'll check
     // the `path` of the event and trigger corresponding callbacks
-    this.components.forEach(component => {
-      const selection = select(component.element)
+    // We also attach to the container itself: a gap between triggers (e.g. between bars) can be
+    // covered by a sibling component's element, which the per-component listener below never sees
+    const elements = [...this.components.map(c => c.element), this._container]
+    elements.forEach(element => {
+      const selection = select(element)
       selection
         .on('mousemove.tooltip', (e: MouseEvent) => {
           const { config: currentConfig } = this // get latest config because it could have been changed after the event was triggered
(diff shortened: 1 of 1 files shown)


rokotyan:
@Vincentdevreede Thanks!

@lee00678 @suryahanumandla I did some investigations: while it looks like a hack, this seems to be a reasonable fix. We would like to test it properly before merging this. I'll do it on my end and if you could try it in your apps, that would be great.

lee00678:
> @Vincentdevreede Thanks!
> 
> @lee00678 @suryahanumandla I did some investigations: while it looks like a hack, this seems to be a reasonable fix. We would like to test it properly before merging this. I'll do it on my end and if you could try it in your apps, that would be great.

Noted, I will do some testing as well.

lee00678:
@Vincentdevreede I think this update breaks `allowHover` 

This is the behavior on `main` 

https://github.com/user-attachments/assets/dd9dac07-f708-4d7c-9b37-7e09d339f31a


This is your branch 

https://github.com/user-attachments/assets/ec640cb0-6f88-4678-abf6-afa8acc5dd67

The new listener attaches to the container in https://github.com/f5/unovis/pull/859/changes#diff-349d9b2f94ecdc1fe4615a4979bf0fee791c4006b0fed078fd416c31f90f0eb1R353, and on any non-trigger mousemove it calls hide when shown in https://github.com/f5/unovis/pull/859/changes#diff-349d9b2f94ecdc1fe4615a4979bf0fee791c4006b0fed078fd416c31f90f0eb1R405.

cc @rokotyan

rokotyan:
@lee00678 Confirmed, and the root cause is exactly the two lines you linked: the new container-level `mousemove` listener also receives events bubbling up from the tooltip `div` itself, because the tooltip is appended inside the container. With `allowHover: true` the tooltip is interactive (no `pointer-events: none`), so hovering it triggers the container listener, which finds no trigger in the event path and calls `hide()` — and with `followCursor` the new re-placement block also drags the tooltip around under the cursor.

The fix is a small guard at the top of the `mousemove` handler, right after `path` is computed — events originating from the tooltip element are owned by its dedicated `mouseenter`/`mouseleave` hover listeners:

```ts
// Events coming from the tooltip itself (possible when `allowHover` is enabled, because the tooltip
// lives inside the container) are handled by the tooltip's own hover listeners, see below
if (path.includes(this.element)) return
```

This is safe for `allowHover: false` because the tooltip then has `pointer-events: none` and can never appear in an event path, so the gap-freeze and fade-in fixes from this PR are unaffected.

Verified against the "Tooltip Sticky Position" dev example: without the guard the tooltip disappears as soon as the cursor moves onto it; with the guard it stays shown and remains interactive, and the tooltip's own `mouseleave` still hides it.

lee00678:
@Vincentdevreede you want to add the fix in so we can have this PR merged?

Vincentdevreede:
@lee00678 @rokotyan Updated the branch 👍

lee00678 (approved):
Fix looks good! Thanks for the contribution.

PR #860, opened 2026-07-24
Patch Release 1.6.8 from branch 1.6.7

lee00678:
This is a release branch for a patch (1.6.8)

Cherry pick https://github.com/f5/unovis/pull/857 into 1.6.7 branch.

Removed `packages/vue/src/components/boxplot/index.vue` as the component is not available yet.

I don't think we should merge this into `main`

Have not released yet.


suryahanumandla:
@lee00678 It looks like the changes were pulled from the 1.6.6 branch instead of the 1.6.7 branch.

lee00678:
> @lee00678 It looks like the changes were pulled from the 1.6.6 branch instead of the 1.6.7 branch.

Yes, I was wondering about that. But use the release/1.6.7 branch which is the only 1.6.7 branch I see. Can you check? I think you released 1.6.7?

PR #868, opened 2026-08-04
Component | Graph: Add configurable node label and sub-label wrapping

lee00678:
Our graph node does not support label wrapping, this PR adds it to label and sublabels. Following the style of sankey and axis, also added the option to force word break. 

## Screenshots / recordings
Screenshot 2026-08-04 at 1 57 15 PM

### DIFF
--- a/packages/angular/src/components/graph/graph.component.ts
+++ b/packages/angular/src/components/graph/graph.component.ts
@@ -297,11 +297,23 @@ export class VisGraphComponent<N extends GraphInputNode, L extends GraphInputLin
   /** Defines whether to trim the node labels or not. Default: `true` */
   @Input() nodeLabelTrim?: BooleanAccessor<N>
 
+  /** Defines whether to wrap the node labels or not. When enabled, wrapping takes precedence over trimming. Default: `false` */
+  @Input() nodeLabelWrap?: BooleanAccessor<N>
+
   /** Node label trimming mode. Default: `TrimMode.Middle` */
   @Input() nodeLabelTrimMode?: GenericAccessor<TrimMode | string, N>
 
   /** Node label maximum allowed text length above which the label will be trimmed. Default: `15` */
   @Input() nodeLabelTrimLength?: NumericAccessor<N>
+
+  /** Maximum width of the wrapped node label in pixels. When not set, wrapped labels use a width derived from the node size. Default: `undefined` */
+  @Input() nodeLabelWidth?: NumericAccessor<N>
+
+  /** Node label text wrapping separator. String or array of strings. Default: `undefined` */
+  @Input() nodeLabelWrapSeparator?: string | string[]
+
+  /** Force word break for node labels when they don't fit. Default: `false` */
+  @Input() nodeLabelForceWordBreak?: boolean
 
   /** Node sub-label accessor function or constant value: Default: `''` */
   @Input() nodeSubLabel?: StringAccessor<N>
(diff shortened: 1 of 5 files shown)


rokotyan:
@lee00678 Should we try using UnovisText for this? It supports quite extensive formatting.

lee00678:
> @lee00678 Should we try using UnovisText for this? It supports quite extensive formatting.

I think this sort of uses it via `renderTextToSvgTextElement`. I didn't really change the type of nodeLabel because I think it's a bigger update. We could consider that in a follow up PR?

suryahanumandla (approved):
LGTM!

rokotyan:
@lee00678 A few good points from an agentic review (Fable 5):

## Blocking issue: broken `setLabelRect` callers

`setLabelRect` went from 3 parameters to 1, but [[panel/index.ts:120](https://claude.ai/epitaxy/packages/ts/src/components/graph/modules/panel/index.ts:120)](packages/ts/src/components/graph/modules/panel/index.ts:120) and [[panel/index.ts:126](https://claude.ai/epitaxy/packages/ts/src/components/graph/modules/panel/index.ts:126)](packages/ts/src/components/graph/modules/panel/index.ts:126) still call it with 3 arguments. I applied the PR locally and ran `tsc` — it fails with `TS2554: Expected 1 arguments, but got 3` in both places. CI is green only because the build pipeline doesn't run a strict type-check over these files. It happens to work at runtime (extra args are ignored and the new implementation handles the panel's single-text case), but the panel calls must be updated.

## API design: this invents a third wrap/trim pattern

The codebase already has a convention for "wrap vs. trim": `FitMode` (`Wrap | Trim`) — used by Axis (`tickTextFitMode`) and Sankey (`labelFit`). This PR instead adds a boolean `nodeLabelWrap` with a documented precedence rule ("wrapping takes precedence over trimming"). Using `nodeLabelFitMode?: GenericAccessor<FitMode | string, N>` would:
- match the existing convention,
- make the precedence caveat unnecessary (the mode is mutually exclusive by construction),
- drop 2 of the 12 new config props (`nodeLabelWrap`, `nodeSubLabelWrap`).

Also, `nodeLabelWidth` is misleadingly generic — it only applies when wrapping. Either follow Axis (`tickTextWidth` applies to both wrap *and* trim) or name it `nodeLabelWrapWidth`. And `nodeLabelWrapSeparator` vs. Axis's `tickTextSeparator` / Sankey's `labelTextSeparator` — dropping "Wrap" would match the siblings.

## Performance

`updateNodes` now unconditionally calls `getBBox()` (forced layout/reflow) plus an extra `getComputedStyle` (sub-label font size) **per node, per update** — and `updateNodes` runs on throttled zoom updates. Previously `getBBox` only happened on hover or at zoom ≥ Level3. On large graphs this is a real regression. Cheap fixes:

- Skip the measurement when there's no sub-label text (the common case) — it's only needed to position the sub-label under the label.
- For the non-wrapped case the label height is effectively `fontSize`; for wrapped text, `getWrappedText` + `estimateWrappedTextHeight` (already exported from [[utils/text.ts:487](https://claude.ai/epitaxy/packages/ts/src/utils/text.ts:487)](packages/ts/src/utils/text.ts:487)) estimates height without touching layout.

## Behavior / compatibility notes

- **DOM structure change is user-visible:** the sub-label is now a sibling `<text>` instead of a `<tspan>` inside the label text, and `labelTextContent` moved from the tspan to the text element. Anyone styling via `Graph.selectors` descendant patterns may break. Worth a line in the PR description/changelog.
- The sub-label offset change (measured label height instead of fixed `dy: 1.1em`) is an improvement — sub-labels without labels no longer float below empty space.
- On hover with wrap enabled, `renderNodeLabelText(…, true)` re-renders identical content (wrap ignores trim, so expand is a no-op) — harmless, but an early return would avoid re-wrapping on every mouseenter.

## Verdict

The feature itself is reasonable and follows the Axis/Sankey wrapping utilities, but it needs a revision before merge: **(1)** fix the `setLabelRect` call sites in the panel module (real type error), **(2)** reconsider `nodeLabelWrap` in favor of the existing `FitMode` convention before this API ships and freezes, **(3)** deduplicate `renderNodeLabelText` and simplify the bbox union — your instinct is right that roughly half of the new logic in `node/index.ts` collapses into a shared per-text helper, and **(4)** gate the new per-node `getBBox` to avoid a zoom-time perf regression.

lee00678:
Closing in favor of #879

PR #869, opened 2026-08-04
Misc | Dependency: Replace rollup with vite

lee00678:
## What & why

This PR attempts to replace rollup with vite. 
Here's the key updates:

Package | Build tool | Status
-- | -- | --
@unovis/ts | Vite (was Rollup) | ✅ Migrated
@unovis/react | Vite (was Rollup) | ✅ Migrated
@unovis/vue | Vite | ✅ Already Vite
@unovis/solid | Vite | ✅ Already Vite
@unovis/svelte | @sveltejs/package CLI | ✅ No change needed
@unovis/angular | Angular CLI (ng build) | ✅ No change needed
react autogen script | tsx (was Rollup) | ✅ Migrated
vue autogen script | tsx (was Rollup) | ✅ Migrated
svelte autogen script | tsx (was Rollup) | ✅ Migrated
angular autogen script | tsx (was Rollup) | ✅ Migrated
svelte gallery dev server | Rollup | ⚠️ Kept — requires Svelte 3→4 upgrade first

Notable source changes:

[style.ts](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) — CSS import changed to ?inline (Vite idiom for CSS-as-string)
[declaration.d.ts](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) — added [*.css?inline](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) type declaration

I think this is a good first step as the svelte was only used for gallery and we already have `pnpm dev:gallery` now, even not with a full dev check list for svelte we could get by. 

There is also an effort of upgrade svelte, so hopefully once that's done we could do a full migrate. 

## Checklist

- [x] Dev examples
- [x] Dev gallery examples (`pnpm dev:gallery`)
- [x] Gallery example
- [ ] Docs

## Screenshots / recordings

cmds-report passing after a clean install.
Screenshot 2026-08-05 at 9 53 44 AM


suryahanumandla (dismissed):
LGTM!

rokotyan:
**JSX runtime change narrows React compatibility.**
The old build emitted classic React.createElement (tsconfig has "jsx": "react"); the new output imports react/jsx-runtime, which only exists in React ≥16.14. The peer range still advertises react: ">=16.8.0". Either set jsxRuntime: 'classic' in @vitejs/plugin-react, or bump the peer floor to >=16.14. (Also worth knowing: for a build-only setup, esbuild handles TSX by itself — the plugin is optional.)

**Published JS is now identifier-minified.** Vite lib mode defaults to esbuild minification, so dist code has single-letter identifiers where the old output was fully readable. Exports and sourcemaps are intact (I verified the API surface is identical), but if readable published code is something you value for debugging in node_modules, set build.minify: false in both configs.

**Redundant but harmless:** react's build script runs rimraf dist and the config also sets emptyOutDir: true.

**Major concern: React TimeSeries styles are silently lost**  — _I don't think anyone uses this so it might not be an issue_
The published 1.6.7 package injects the composite's CSS at import time — composites/time-series/style.module.css.js contains the CSS string and calls styleInject(css) (via rollup-plugin-postcss's default inject). With the new config, I built the PR branch and got:

dist/composites/time-series/style.module.css.js now exports only the hashed class-name map — no CSS, no injection.
The actual rules are extracted to dist/react.css, an orphan file nothing imports and package.json doesn't reference.
So VisTimeSeries will render with class names that match no rules — a silent visual break that no build or CI step catches. The repo already has the fix pattern in-house: packages/vue/vite.config.ts uses vite-plugin-css-injected-by-js for exactly this Vite lib-mode behavior. Options, roughly in order of preference:

Add vite-plugin-css-injected-by-js to packages/react/vite.config.ts (matches the Vue precedent). Caveat for both packages: "sideEffects": false in package.json means aggressive bundlers may drop the injection module anyway — worth listing the injecting file in sideEffects while you're here.
Or migrate the composite's three rules to @emotion/css, consistent with how the rest of unovis styles itself, and delete the .module.css entirely.
Either way, dist/react.css shouldn't ship unreferenced.

---

**Separately**, I noticed a pre-existing regression while comparing exports (not caused by this PR): estimateStringPixelLength and getPreciseStringLengthPx were public in 1.6.7 but got dropped from the barrel when they moved to utils/text-measure.ts — I've flagged that as a separate task chip.

lee00678:
> **JSX runtime change narrows React compatibility.** The old build emitted classic React.createElement (tsconfig has "jsx": "react"); the new output imports react/jsx-runtime, which only exists in React ≥16.14. The peer range still advertises react: ">=16.8.0". Either set jsxRuntime: 'classic' in @vitejs/plugin-react, or bump the peer floor to >=16.14. (Also worth knowing: for a build-only setup, esbuild handles TSX by itself — the plugin is optional.)
> 
> **Published JS is now identifier-minified.** Vite lib mode defaults to esbuild minification, so dist code has single-letter identifiers where the old output was fully readable. Exports and sourcemaps are intact (I verified the API surface is identical), but if readable published code is something you value for debugging in node_modules, set build.minify: false in both configs.
> 
> **Redundant but harmless:** react's build script runs rimraf dist and the config also sets emptyOutDir: true.
> 
> **Major concern: React TimeSeries styles are silently lost** — _I don't think anyone uses this so it might not be an issue_ The published 1.6.7 package injects the composite's CSS at import time — composites/time-series/style.module.css.js contains the CSS string and calls styleInject(css) (via rollup-plugin-postcss's default inject). With the new config, I built the PR branch and got:
> 
> dist/composites/time-series/style.module.css.js now exports only the hashed class-name map — no CSS, no injection. The actual rules are extracted to dist/react.css, an orphan file nothing imports and package.json doesn't reference. So VisTimeSeries will render with class names that match no rules — a silent visual break that no build or CI step catches. The repo already has the fix pattern in-house: packages/vue/vite.config.ts uses vite-plugin-css-injected-by-js for exactly this Vite lib-mode behavior. Options, roughly in order of preference:
> 
> Add vite-plugin-css-injected-by-js to packages/react/vite.config.ts (matches the Vue precedent). Caveat for both packages: "sideEffects": false in package.json means aggressive bundlers may drop the injection module anyway — worth listing the injecting file in sideEffects while you're here. Or migrate the composite's three rules to @emotion/css, consistent with how the rest of unovis styles itself, and delete the .module.css entirely. Either way, dist/react.css shouldn't ship unreferenced.
> 
> **Separately**, I noticed a pre-existing regression while comparing exports (not caused by this PR): estimateStringPixelLength and getPreciseStringLengthPx were public in 1.6.7 but got dropped from the barrel when they moved to utils/text-measure.ts — I've flagged that as a separate task chip.

All issues here have be resolved in separated commits. cc @rokotyan

suryahanumandla:
LGTM!

PR #870, opened 2026-08-04
Adaptive tick sets

devgru:
## What & why

This PR introduces adaptive tick sets, a new axis option, that picks tick labels to display in a coordinated manner, instead of filtering them one by one.

## Checklist

- [x] Dev examples
- [x] Docs

## Screenshots / recordings

https://github.com/user-attachments/assets/390763b8-971d-4e13-8f8b-fe5a966aa7a1

https://github.com/user-attachments/assets/4fa28497-e76a-4ea6-a5c2-20db094c89f6

### DIFF
--- a/packages/angular/src/components/axis/axis.component.ts
+++ b/packages/angular/src/components/axis/axis.component.ts
@@ -96,9 +96,6 @@ export class VisAxisComponent<Datum> implements AxisConfigInterface<Datum>, Afte
   /** Label text trim mode: `TrimMode.Start`, `TrimMode.Middle` or `TrimMode.End`. Default: `TrimMode.Middle` */
   @Input() labelTextTrimType?: TrimMode | `${TrimMode}`
 
-  /** Label text separators for wrapping. Default: `[' ', '-']` */
-  @Input() labelTextSeparator?: string[]
-
   /** Font color of the axis label as CSS string. Default: `null` */
   @Input() labelColor?: string | null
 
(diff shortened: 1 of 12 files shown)


packages/angular/src/components/axis/axis.component.ts
```diff
@@ -96,6 +96,9 @@ export class VisAxisComponent<Datum> implements AxisConfigInterface<Datum>, Afte
   /** Label text trim mode: `TrimMode.Start`, `TrimMode.Middle` or `TrimMode.End`. Default: `TrimMode.Middle` */
   @Input() labelTextTrimType?: TrimMode | `${TrimMode}`
 
+  /** Label text separators for wrapping. Default: `[' ', '-']` */
+  @Input() labelTextSeparator?: string[]
```
lee00678: I think this somehow got sneak in? I don't think this config belongs to axis, it's part of sankey.

packages/ts/src/components/axis/index.ts
```diff
@@ -346,25 +377,148 @@ export class Axis<Datum> extends XYComponentCore<Datum, AxisConfigInterface<Datu
...
+      const localRect = { x: x0, y: y0, width, height }
+      const rect = angleRad ? getRotatedRectAabb(localRect, angleRad) : localRect
+      return { ...rect, x: tickPosition[0] + rect.x, y: tickPosition[1] + rect.y }
+    })
+  }
+
+  /** Returns the cached tick label text style, resolving it on first use — from the given
+   * element when provided, otherwise from a throwaway element with the same classes */
+  private _getTickTextStyle (textElement?: SVGTextElement): TickTextStyle {
+    if (this._tickTextStyleCached) {
```
lee00678, line 496: I know _tickTextStyleCached is not introduced in this PR, but it seems this cache is never cleared. 

If [tickTextFontSize] changes between renders (e.g., a [setConfig], the fitting logic will predict label widths with the stale font metrics and may pick the wrong tick set.

packages/angular/src/components/axis/axis.component.ts
devgru: good point! it did, indeed.

AFAIK Nikita ported the concept from Sankey diagram to the Axis, but we certainly don’t need it in this PR.

packages/ts/src/components/axis/index.ts
devgru: fair point, please take a look at the proposed solution in the latest commit

packages/angular/src/components/axis/axis.component.ts
rokotyan: @devgru @lee00678 The PR I forgot to create https://github.com/f5/unovis/pull/872

lee00678 (approved):
LG, thanks for updating this!

PR #872, opened 2026-08-06
Component | Axis: Config for label trim separator character

rokotyan:
## What & why

Adding a new config property: `labelTextSeparator`. 
@lee00678, we've been using it for quite a while already, I just forgot to push it to the upstream.

## Checklist

- [x] Dev examples
- [x] Docs

## Screenshots / recordings

https://github.com/user-attachments/assets/c5da40d9-0697-41b5-a753-5edb94fb5e22

### DIFF
--- a/packages/ts/src/components/axis/config.ts
+++ b/packages/ts/src/components/axis/config.ts
@@ -23,6 +23,8 @@ export interface AxisConfigInterface<Datum> extends Partial<XYComponentConfigInt
   labelTextFitMode?: FitMode | `${FitMode}`;
   /** Label text trim mode: `TrimMode.Start`, `TrimMode.Middle` or `TrimMode.End`. Default: `TrimMode.Middle` */
   labelTextTrimType?: TrimMode | `${TrimMode}`;
+  /** Label text wrapping separator. String or array of strings. Default: `undefined` */
+  labelTextSeparator?: string | string[];
   /** Font color of the axis label as CSS string. Default: `null` */
   labelColor?: string | null;
   /** Sets whether to draw the grid lines or not. Default: `true` */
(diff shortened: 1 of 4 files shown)


packages/ts/src/components/axis/config.ts
```diff
@@ -23,6 +23,8 @@ export interface AxisConfigInterface<Datum> extends Partial<XYComponentConfigInt
   labelTextFitMode?: FitMode | `${FitMode}`;
   /** Label text trim mode: `TrimMode.Start`, `TrimMode.Middle` or `TrimMode.End`. Default: `TrimMode.Middle` */
   labelTextTrimType?: TrimMode | `${TrimMode}`;
+  /** Label text separators for wrapping. Default: `[' ', '-']` */
+  labelTextSeparator?: string[];
```
lee00678: The default says `[' ', '-']`, but it's actually undefined right? Should we only allow `string[]`? for `tickTextSeparator?: string | string[];`

And for wrapSVGText we have `[' ', '-', '.', ',']` as defaults, should we keep the default the same for consistancy?

packages/website/docs/auxiliary/Axis.mdx
```diff
@@ -69,6 +69,23 @@ Change the label color with the `labelColor` property.
 The spacing between the label and the axis itself can be set with the `labelMargin` property:
 <XYWrapperWithInput {...axisProps()} label="Label" inputType="range" inputProps={{ min: 0, max: 50, step: 1}} defaultValue={5} property="labelMargin"/>
 
+### Label Text Separator
+When a label is too long to fit, _Axis_ wraps it onto multiple lines (this happens when `labelTextFitMode` is
+set to `FitMode.Wrap`, which is the default). You can control where the text is allowed to break by providing the
+`labelTextSeparator` property with a `string` or `string[]` of separators. By default the label breaks on spaces and hyphens.
```
lee00678: It actually says `string` or `string[]` here as well.

packages/ts/src/components/axis/config.ts
rokotyan: @lee00678 I'm glad you spotted this. Fixed.

lee00678 (approved):

PR #873, opened 2026-08-06
MCP & SSR: Add @unovis/ssr and @unovis/mcp, render Unovis charts headlessly

rokotyan:
_Unovis_ can't be driven by an AI agent today, and it can't render outside a browser at all. This adds `@unovis/mcp` — an [MCP](https://modelcontextprotocol.io) server that turns data into _Unovis_ charts and hands back an SVG, a PNG, a live HTML file, or ready-to-paste component source. Everything renders **locally in Node**: no browser download, no remote rendering service, and the data never leaves the machine.

```bash
claude mcp add unovis -- npx -y @unovis/mcp
```

> [!NOTE]
> Marked draft until the stable `@unovis/ts` release ships the core fixes below. Prereleases are live for early adopters (`@unovis/ts@1.7.0-mcp.x`, `@unovis/ssr@1.7.0-mcp.x`, `@unovis/mcp@1.7.0-mcp.x`).

## What an agent can generate

Sixteen tools — fifteen chart types plus `get_unovis_info` for cheap capability discovery:

| Family | Tools |
|---|---|
| XY | `generate_line_chart`, `generate_area_chart`, `generate_bar_chart`, `generate_scatter_plot`, `generate_timeline_chart`, `generate_boxplot` |
| Radial | `generate_donut_chart`, `generate_nested_donut_chart`, `generate_radial_bar_chart` |
| Hierarchy & flow | `generate_sankey_diagram`, `generate_treemap`, `generate_chord_diagram`, `generate_network_graph` |
| Grid & geo | `generate_heatmap`, `generate_choropleth_map` |

Every tool takes flat records plus **field-name accessors** rather than functions — `x: "month"`, `y: ["sales", "cost"]` — which is the shape an LLM reliably produces, and which keeps arbitrary code out of the request:

```json
{
  "name": "generate_line_chart",
  "arguments": {
    "data": [{ "month": "2024-01", "sales": 120, "cost": 80 }, { "month": "2024-02", "sales": 150, "cost": 95 }],
    "x": "month", "xIsTime": true, "y": ["sales", "cost"],
    "seriesLabels": ["Sales", "Cost"], "title": "Revenue"
  }
}
```

Shared across the tools: light/dark theme, title, axis labels, legend, custom palettes, size, a BCP-47 `locale` for date/number formatting, and — for XY charts — `referenceLines`, `referenceBands` and point `annotations`. `generate_network_graph` supports the `force`, `circular`, `concentric` and `dagre` layouts. Field names are validated against the data, so a typo comes back as "unknown field `sale`, available: month, sales, cost" instead of an empty chart.

## Six output types, one spec

`outputType` switches what a tool returns, and every one of them is driven by the same JSON chart spec:

- **`svg`** — a standalone document: no CSS classes, no custom properties, no external references. Drop it in a README, an email, or an ``.
- **`png`** — the same SVG through [resvg](https://github.com/yisibl/resvg-js), with `scale` for 2× output. For clients that can't display SVG.
- **`html`** — one self-contained file with the interactive chart: hover tooltips, a crosshair with a multi-series readout, _Unovis_'s real `BulletLegend`, graph drag/zoom, responsive resize. Nothing external, so it opens offline. The same document doubles as an embeddable widget driven over a small postMessage protocol: a `unovis:ready` handshake reporting `{ version, specVersion }`, re-render, runtime theme switching, and opt-in click events normalized to the caller's own data records — including first-class React Native WebView hosting (JSON strings over the native bridge, `document`-delivered messages on Android). Documents stay small twice over: two prebuilt bundle variants split along the dependency line (`standard` for XY/radial charts, `full` adds the graph/geo/flow engines) with the smallest sufficient one picked per spec, and the chosen bundle travels as a gzip payload behind a ~5kB self-extracting bootstrap — a line-chart file is **138kB** on disk, a graph file 288kB (both were 671kB), with the spec kept as plain text so committed artifacts diff meaningfully.
- **`interactive`** — the widget through the official **MCP Apps extension** (`io.modelcontextprotocol/ui`, SEP-1865): every chart tool declares its template ahead of time so hosts (Claude, ChatGPT, VS Code Copilot, Goose…) can prefetch and security-review it — easy, since the resource ships **empty CSP allowlists** backed by the tested no-network guarantee. The widget implements the full view lifecycle (`ui/initialize` → `initialized` → `tool-result`, host theme adoption, `size-changed`, `ping`), verified against the official AppBridge host implementation in a sandboxed iframe. When a client advertises the capability, **every** result carries the spec in `structuredContent`, so even a plain `svg` call renders a live chart in an Apps host while other clients keep lean responses.
- **`config`** — the chart spec as JSON, for storing, diffing or re-rendering later.
- **`code`** — component source for `ts`, `react`, `svelte`, `vue`, `angular` or `solid`, so "make me a chart" ends in code the developer owns rather than an image they can't edit.

The single-spec design is the point: the headless renderer, the browser widget and the code generator all consume the same IR, so a chart can't look one way as an SVG and another way in the app.

## Headless rendering

_Unovis_ is browser-first — it builds SVG by mutating a live DOM with D3, measures text with canvas, resolves its theme through `--vis-*` custom properties, and schedules every render through `requestAnimationFrame`. `src/env/` provides exactly those APIs on top of jsdom, installed **before** `@unovis/ts` is imported because the library captures its environment at module load:

- a **flushable rAF queue**, so rendering is synchronous and deterministic instead of spread over frames;
- **real canvas metrics** (`@napi-rs/canvas`) for `measureText`, which is what drives label trimming, wrapping and axis auto-margins;
- **geometry polyfills** — `getBBox`, `getBoundingClientRect`, `getComputedTextLength`, `getTotalLength`/`getPointAtLength` — since jsdom performs no layout at all: text boxes from canvas metrics, shapes from attribute math, paths via `svg-path-bounds`/`svg-path-properties`, groups from the union of transformed children;
- a **`getComputedStyle` wrapper** that resolves the `--vis-*` cascade jsdom doesn't implement.

Determinism comes from forcing `duration: 0` everywhere (`smartTransition` then returns the plain selection), draining the frame queue until it stays empty, and waiting for the container's `onRenderComplete` — with a separate signal for components whose layout is async. If either signal never arrives the render throws instead of emitting a half-drawn chart.

Text metrics are pinned rather than left to the host: Inter 4.1 is downloaded once (SHA-256 verified) into `~/.cache/unovis-ssr/fonts/` and registered with the canvas, so measurement matches the font the output declares. `UNOVIS_SSR_NO_DOWNLOAD=1` and `UNOVIS_SSR_FONTS_DIR` opt out (the `UNOVIS_MCP_*` names still work).

Making the SVG standalone is its own pass: import external defs, inline emotion's rules from the CSSOM, bake `var(--vis-*)` to literals for the requested theme, normalize `url(http://localhost/#id)` references, rewrite random guids to `prefix-N` so several charts can share a document, and synthesize the title/legend header that _Unovis_ containers have no concept of.

## It's also a library — two, in fact

Headless rendering ships as its own package, **`@unovis/ssr`**: the jsdom environment and shims, the standalone-SVG post-processor, `renderToSvg` and `svgToPng` — SSR for charts, without an agent (or the MCP surface) anywhere in the picture. `@unovis/mcp` consumes it exactly like any other caller would and re-exports the API, so both import paths work:

```ts
import { renderToSvg } from '@unovis/ssr'

const { svg } = await renderToSvg({ width: 800, height: 400, theme: 'dark' }, ctx => {
  const line = new ctx.unovis.Line({ x: d => d.x, y: d => d.y })
  const container = new ctx.unovis.XYContainer(ctx.container, {
    components: [line], width: ctx.width, height: ctx.height,
    duration: 0, onRenderComplete: ctx.onRenderComplete,
  }, data)
  return { root: container.svg.node(), destroy: () => container.destroy() }
})
```

The spec itself is a versioned contract (`specVersion`, refused with an explicit error when newer than the renderer), and two dependency-light subpaths serve consumers that only build or validate specs: `@unovis/mcp/recipes` (recipes plus their spec-building helpers — accessors, axis defaults, calendar-aligned time ticks) and `@unovis/mcp/spec` (the IR types). `renderToSvg` takes _any_ hand-built _Unovis_ chart, not just ones the tools can describe — chart-in-CI, chart-in-email, chart-in-PDF all work from this one function. `renderChart` renders a spec, `svgToPng` rasterizes, `buildChartDocument` produces the interactive file, `generateCode` emits framework source, and the tool registry is exported so these tools can be mounted inside someone else's MCP server.

## Core changes

Four small fixes belong in `@unovis/ts` rather than in shims, and are made there:

- **`ContainerCore`** falls back to the configured width/height when `clientWidth` is 0. Detached and non-layouting environments previously sized every chart to zero, which is also why axes rendered without ticks.
- **`getPixelsPerInch`** pre-seeds its cache before measuring. Without layout, `getComputedStyle` returns the specified value (`'128in'`) and the function recursed until the stack ran out.
- **Root `index.js` / `maps.js` entries** so `@unovis/ts` and `@unovis/ts/maps` resolve under a standards-compliant Node loader (excluded from the dts type-check program so the entry-declaration guard from #890's fix stays green).
- **`Graph`: a `:not()` link selector was missing its closing parenthesis.** CSS parsing auto-closes unbalanced parens at end of input, so browsers forgave it — jsdom's selector engine correctly rejects it, which made every headless graph render throw after layout completed.

(An earlier commit here also restored the `dist` entry declarations the vite migration silently dropped — that fix has since landed upstream properly, with diagnostics cleared and an emission guard, so this branch dropped its workaround. Filed as [#890](https://github.com/f5/unovis/issues/890), alongside [#888](https://github.com/f5/unovis/issues/888) — text measurement via canvas instead of getBBox — and [#889](https://github.com/f5/unovis/issues/889) — event binding without the 500ms throttle window — the two remaining issues that shrink the shim layer.)

Plus a dependency bump to `@unovis/dagre-layout@0.8.8-3` and `@unovis/graphlibrary@2.2.0-3`, whose earlier releases shipped extensionless ESM imports that Node can't resolve — that's what unblocks the `dagre` graph layout here.

The trend matters more than the diff: as the core gets SSR-friendlier, the shim layer shrinks.

## Hardened for hostile input

- **File writes are gated on the HTTP transport.** `outputPath` on an unauthenticated network endpoint was a remote file-write primitive; writes are now off by default on HTTP and opt in with `--allow-write-dir <dir>` (resolved-path prefix checks; `..` and sibling-prefix tricks covered). stdio keeps its user's own authority.
- **Every string surface is fuzzed** — titles, axis labels, legend names, field names, data values — with parse-based assertions (no script elements, no `on*` attributes) across SVG output, executed documents, and tooltip templates. The suite immediately found and fixed a real hole: a string `width` in a hand-built spec could escape the document's `<style>` block into a script element.
- **The chart spec is a published, enforced contract**: `specVersion` gated at render, a generated JSON Schema shipped as `@unovis/mcp/chart-spec.schema.json` for host-side validation, and a frozen baseline in CI that fails on any non-additive change — removed property, changed type, lost enum value, new requirement.

## Included

- **`packages/ssr`** — `@unovis/ssr`: the headless environment (jsdom + shims), standalone-SVG post-processing, `renderToSvg`, `svgToPng`, font provisioning. Moved with history from the MCP package; the shim installers are exported individually for consumers hosting Unovis in their own jsdom.
- **`packages/mcp`** — the server: renderer, spec materializer, 15 recipes, codegen, browser widget (two size variants + compression), CLI (stdio by default, streamable HTTP behind `--transport http`), and per-deployment tool filtering via `--tools` / `DISABLED_TOOLS`.
- **`packages/ts`** — the fixes above and the dependency bump.
- **Website** — an _MCP Server_ section at [`/docs/mcp`](https://unovis.dev/docs/mcp) (10 pages, including embedding in native WebViews), generated from `packages/mcp/docs` by `scripts/sync-mcp-docs.mjs` so the npm tarball keeps its own copy and the two can't drift; the `mcp` CI job re-runs the sync and fails on a diff. Plus a pointer from _Introduction_.
- **Commitlint** — `MCP` and `SSR` join the commit types (scope optional, like `Release`/`CI`).
- **CI** — a new job covering both packages: build (ts → ssr → mcp), type-check, lint, the full test suites, the docs-sync check, a **Chromium browser lane** (every family rendered with real layout and input: hover → crosshair readout, click → embed event, theme flip, and MCP Apps conformance against the official AppBridge host), and `pnpm samples`, which renders every fixture in both themes and uploads them as an artifact so a reviewer can look at the charts instead of trusting a green check. Inter is cached, keyed on the file that pins its version and checksum.

## Verification

170 tests across two suites (`@unovis/mcp` 161, `@unovis/ssr` 9) cover the env shims (bbox math, variable resolution, frame flushing), `renderToSvg` driven by hand-written Unovis code, **pixel-based visual regression** per chart type (baselines stored as SVG text, one element per line; both sides rasterize through the same resvg at compare time, so attribute churn passes and a recolored 2px line fails — with the formatter proven pixel-neutral by a zero-diff control), the post-processing passes, the MCP surface over a real SDK client on an in-memory transport (schemas, every output type, error paths, tool filtering, concurrency), the **real widget bundle executed in jsdom** including a simulated hover producing tooltip content, all 15 chart types rendered interactively with a clean console, generated TypeScript **type-checked against `@unovis/ts`**, every tool schema validated against the JSON Schema draft 2020-12 meta-schema (one draft-07 construct took all sixteen tools offline in a real client before this), spec version gating and locale formatting, interaction events round-tripped through a dispatched click on the real bundle, the React Native bridge and theme switching, bundle-variant selection including the explicit error when a spec outruns its bundle, compressed documents executed end-to-end (the whole widget suite runs against them), and an assertion that both widget bundles and the unpack bootstrap contain **no network-capable tokens** — no fetch/XHR/WebSocket, no URLs beyond W3C namespaces — so zero-egress deployments rest on an upstream contract instead of a hand audit.

Beyond the unit suites: the Chromium lane runs on every PR (real layout, paint and input — added after a crosshair bug passed every jsdom test), every fixture is reviewed by eye in both themes via `pnpm samples`, and the tools are driven end-to-end from Claude Code. The MCP Apps lifecycle is verified against the official AppBridge host — which caught that the real initialize params are `appInfo`/`appCapabilities`, not the shape secondary sources document.

Graph charts are deliberately **not** snapshotted. They keep settling for a few frames after their layout reports completion (fit-view, label collision), and `force` additionally calls `Math.random()`, so exact geometry depends on machine load. The renderer waits out a short grace period so real output doesn't lose those late frames, but a snapshot would be a false signal that fails randomly under CI load. Structural assertions, the widget matrix and an async-layout probe cover them instead.

## Notes and known limits

- **The `elk` graph layout is not offered.** It works through a hand-written spec in Node, but its 1.4MB engine loads via dynamic import and would have to be inlined into every generated HTML file — so offering it as a tool option would mean `outputType: "html"` silently producing a broken chart.
- **`outputType: "interactive"`** rides the official MCP Apps extension; host adoption is broad but still spreading, and behavior in each commercial host isn't individually tested — `html` remains the always-works path.
- **The `tsx` loader can't load the graph layouts'** dynamic imports — the layout promise never settles. `pnpm dev` and `pnpm samples` therefore build first and run the compiled output; plain Node and vitest are both fine.
- **Visual baselines are font-dependent.** Both comparison sides rasterize on the same machine with pinned Inter, so the tight pixel tolerance only absorbs sub-pixel jitter; a disagreement beyond it means font provisioning broke, not the renderer.
- **Data passes through the model's context**, so hundreds of rows are comfortable and tens of thousands are wasteful — aggregate first, or use the library directly, where data never touches a model.

Follow-ups I'd keep out of this PR: a combined bar+line tool and a real-browser (Cypress) test for the interactive output.


lee00678:
This is awesome. We have feature that uses download svg as well.

rokotyan:
@lee00678 @suryahanumandla (cc @devgru) The initial test went well. Asked Claude to create an dashboard artifact with Unovis charts
  
https://github.com/user-attachments/assets/e33fd222-8fd6-4a7a-826a-4c396d30ef53

rokotyan:
Also asked it to render the same charts as PNG / SVG. We might need to tweak some defaults but rendering seems to be working well.

01-daily-volume-by-severity
02-severity-mix
03-days-past-sla
04-source-to-verdict-sankey
05-signal-to-noise-by-source
06-mitre-tactic-x-severity
07-top-mitre-techniques
08-arrival-hour

lee00678:
Yea this looks great. I assume dark mode would also out of the gate since it's just taking a snapshot of the same displayed svg?

rokotyan:
@lee00678 Marking this a ready for review, but if you prefer not to include it into 1.7, I'm fine with that. Maybe we should spend more time now refining the recipes, so we introduce less schema changes in the future.

On the other hand, it barely touches the main bundle and contains a buch of useful features

lee00678 (dismissed):
Awesome addition, can't wait to use the SVG export feature.

rokotyan:
## CodeQL: the remaining 12 alerts are false positives

All 12 are XSS findings (`js/xss` and `js/xss-through-dom`) on six `appendChild` calls in `@unovis/ts`:

| Location | Call |
|---|---|
| `containers/xy-container/index.ts:152, 156, 177, 183` | `this.element.appendChild(xAxis / yAxis / crosshair / annotations .element)` |
| `containers/single-container/index.ts:55, 65` | `this.element.appendChild(component / annotations .element)` |
| `components/tooltip/index.ts:84` | `this._container.appendChild(this.element)` |

Lines 177, 183, 55, 65 and 84 are reported by both queries, hence 12 alerts for 6 sites.

### Why they fire

CodeQL's XSS queries are taint-tracking: sources are untrusted inputs, sinks are HTML-interpreting operations. This PR introduces two sources — `event.data` from `postMessage` and `specEl.textContent` from the JSON embedded in the widget document — so the parsed chart spec is tainted. Taint then propagates coarsely: an object built from tainted values is tainted, and any property read from a tainted object is tainted. So `spec → materializeChart → new XYContainer(containerConfig) → containerConfig.xAxis → .element` keeps the taint all the way to the argument of `appendChild`, which CodeQL's DOM model treats as an HTML-insertion sink.

None of these files change in this PR, and `main` is clean, because `main` has no untrusted source feeding these paths.

### Why they're false positives

`appendChild` takes a `Node`, not a string. The argument in every case is an element the library created via `createElement` / `createElementNS`; there is no parse step that could turn spec content into markup on these lines. Spec strings reach the DOM downstream only through `.text()` / `textContent` on labels, which is safe by construction. The analyzer isn't wrong about the *taint* — the spec is untrusted — it's wrong about the *sink*: appending a library-created node is a container operation, not an injection point.

### Why not fix in code

There is no code-level change that removes them honestly. Rewriting the node insertion or excluding the widget entry from analysis would suppress the alerts without changing any behavior, and the latter would also hide real paths. Dismissing the 12 as **False positive** in the Security tab is the correct disposition; dismissals are fingerprinted, so they persist across pushes and future PRs.

### Contrast with what was real

The earlier alerts on the same taint path were genuine sinks and are fixed and tested in this PR: d3 `.html()` on a spec-controlled node shape string (`Graph`), function calls through spec-controlled property names (`Plotband` / `Plotline` label positions, and the MCP renderer's component / projection / event-extractor lookups), plus the ReDoS and error-markup findings in `@unovis/ssr` and `@unovis/mcp`. The 12 left are the containers that carry the already-safe elements around.

lee00678:
Confirming the 12 alerts are false positive.

lee00678 (approved):
this is an amazing addition.

PR #875, opened 2026-08-06
React: Render when a component's config changes

rokotyan:
Fixes a React repaint bug: updating a component-level prop (e.g. a `color` accessor) produced no repaint when the new value reached the component through React context or a parent's state — the container's props stayed equal, its `React.memo` bailed out, and nothing ever rendered. The chart stayed one step behind until an unrelated interaction; calling `component.render()` by hand was the only workaround.

The wrapper's effect called `setConfig()`, which only stores the config — containers own the render loop. Now containers publish a `requestRender` callback over React context and every generated component wrapper calls `useContainerRenderOnUpdate()` after updating its config. Requests coalesce into one animation frame and defer to a pending container update, so a change costs exactly one paint; reference-stable props still schedule nothing. This is the React counterpart of the Svelte fix in 8a026a5b (Angular, Vue and Solid already notify their containers).

Also fixes an `isEqual` false positive (own commit): a reference appearing twice as a *sibling* was reported equal without comparing its counterpart, because `visited` kept everything it had ever seen instead of just the recursion stack.

### Before

https://github.com/user-attachments/assets/009b65fd-6c08-49c0-8c97-f0fe45825711

### After

https://github.com/user-attachments/assets/dd1ed9b4-e2fd-46bf-aaf5-4215c634b442

- [x] Core (`isEqual`)
- [x] React containers + autogen + regenerated wrappers
- [x] Dev example (`Heatmap Accessor Update` — live PASS/FAIL readout)

### DIFF
--- a/packages/dev/cypress/e2e/component-accessor-update.cy.ts
+++ b/packages/dev/cypress/e2e/component-accessor-update.cy.ts
@@ -1,0 +1,62 @@
+// Regression test: a component-level config change has to repaint on its own.
+//
+// In this example the selection reaches `VisHeatmap` through React context, so
+// `VisSingleContainer`'s own props stay equal across the update and it never re-renders. That used to
+// mean the component's `setConfig()` stored the new `color` accessor and nothing ever painted it —
+// the chart kept showing the previous selection's colors until some unrelated interaction happened.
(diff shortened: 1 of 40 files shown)


lee00678 (approved):
looks good. cmds-report passes.

PR #877, opened 2026-08-06
Prevent Sankey node height collapse

devgru:
## What & why

A combination of rigid vertical padding and adaptive node height in the Sankey diagram causes the nodes to get a height of 1px, once the amount of nodes is sufficiently large (see attached videos).

This PR introduces a flag that enables adaptive node paddings, allowing to reduce it when the available vertical space is limited, instead of only reducing nodes' height.

## Checklist

- [x] Dev examples
- [x] Docs

### Before:

https://github.com/user-attachments/assets/38ab31ec-b2a2-47bc-a216-96cfcbd547ab

### After:

https://github.com/user-attachments/assets/bbeb4125-903a-475c-9ac0-dbb9c4c8962a


lee00678 (approved):
This looks great!

lee00678 (approved):

PR #879, opened 2026-08-07
Component | Graph: Add configurable node label and sub-label wrapping

rokotyan:
Adds configurable text wrapping for Graph node labels and sub-labels, built on the UnovisText rendering utilities. Supersedes #868 (same feature, reworked implementation) — co-authored with @lee00678, whose PR this builds on.

image

## API

New Graph config options, following the `FitMode` convention already used by Axis and Sankey (instead of a separate wrap boolean with a precedence rule):

- `nodeLabelFitMode` / `nodeSubLabelFitMode` — `FitMode.Trim` (default, current behavior) or `FitMode.Wrap`
- `nodeLabelWidth` / `nodeSubLabelWidth` — wrap width in pixels; falls back to a node-size-derived width
- `nodeLabelSeparator` / `nodeSubLabelSeparator` — wrapping separator(s)
- `nodeLabelForceWordBreak` / `nodeSubLabelForceWordBreak` — hyphenated word break for long words

## Implementation

- **UnovisText utilities extended (non-breaking):**
  - `UnovisText` blocks accept a `className`, so components can keep CSS-variable theming and user CSS overrides on rendered tspans.
  - Per-block `width` / `separator` / `wordBreak` overrides, letting one `<text>` element host blocks with independent wrapping (label + sub-label).
  - `_maxWidth` is now per-block and no longer includes a phantom ellipsis; new `getWrappedTextBounds` helper; `renderTextToSvgTextElement` returns the wrapped blocks it rendered.
- **Graph node labels are now rendered as UnovisText blocks** in a single `renderTextToSvgTextElement` call. Sub-label positioning comes from block layout (no more fixed `1.1em` offset — a sub-label without a label no longer floats below empty space), and no `getBBox` measurements happen on update.
- The DOM stays close to the previous structure: one `<text>` per node with `labelTextContent` / `subLabelTextContent` classed tspans, so existing theming keeps working.
- `setLabelRect` keeps its signature (panel labels unaffected) and now derives the background rectangle from the rendered bbox, which makes it correct for multi-line labels. Note: this repositions label background rects by a couple of pixels vertically compared to the old `-0.64em` heuristic.

## Verification

- `tsc` on `@unovis/ts` introduces no new errors; eslint clean on all touched files.
- Verified in the dev app (`Graph: Node Labels and Sub-labels` example, updated to showcase wrapping): trimming unchanged, multi-line wrapping at configured widths, force word break hyphens, hover expand/restore with background rect resize, wrapped-label rects covering the full text block, CSS-variable colors/fonts intact, panel labels unaffected.

## How to test

Run the updated `Node Labels and Sub-labels — Trimming and Wrapping` example in `packages/dev`; see #868 for the visual reference of the feature.

### DIFF
--- a/packages/ts/src/components/graph/modules/node/index.ts
+++ b/packages/ts/src/components/graph/modules/node/index.ts
@@ -96,7 +96,6 @@ export function createNodes<N extends GraphInputNode, L extends GraphInputLink> 
     label.append('rect').attr('class', nodeSelectors.labelBackground)
     label.append('text')
       .attr('class', nodeSelectors.labelText)
-      .attr('dy', '0.32em')
   })
 }
 
--- a/packages/ts/src/components/graph/modules/node/index.ts
+++ b/packages/ts/src/components/graph/modules/node/index.ts
@@ -227,7 +227,7 @@ export function updateNodes<N extends GraphInputNode, L extends GraphInputLink> 
 ): Selection<SVGGElement, GraphNode<N, L>, SVGGElement, unknown> | Transition<SVGGElement, GraphNode<N, L>, SVGGElement, unknown> {
   const {
     nodeGaugeAnimDuration, nodeStrokeWidth, nodeShape, nodeSize, nodeGaugeValue, nodeGaugeFill,
-    nodeIcon, nodeIconSize, nodeLabel,
+    nodeIcon, nodeIconSize, nodeLabel, nodeSubLabel,
     nodeSideLabels, nodeStroke, nodeFill, nodeBottomIcon,
   } = config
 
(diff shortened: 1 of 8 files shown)


packages/ts/src/components/graph/modules/node/index.ts
```diff
@@ -88,18 +89,14 @@ export function createNodes<N extends GraphInputNode, L extends GraphInputLink>
...
+    // Label. The label and sub-label are rendered into the `<text>` element
+    // as text blocks with the `nodeSelectors.labelTextContent` and
+    // `nodeSelectors.subLabelTextContent` classes, see `renderNodeLabels`
     const label = group.append('g').attr('class', nodeSelectors.label)
     label.append('rect').attr('class', nodeSelectors.labelBackground)
-
-    const labelText = label.append('text')
+    label.append('text')
       .attr('class', nodeSelectors.labelText)
       .attr('dy', '0.32em')
```
lee00678: I don't think we need this anymore. 

`renderTextToSvgTextElement` creates every tspan with an absolute y, so `dy="0.32em"` doesn't feed into the layout calculation anymore.

lee00678:
Noticed a small issue, which may not be introduced by it. 

When there's no label only sublabel, there's no background rendered.

Screenshot 2026-08-07 at 9 52 08 AM

packages/ts/src/components/graph/modules/node/index.ts
rokotyan: Agreed on removing it — the mechanism is slightly different though: `renderTextToSvgTextElement` only sets an absolute `y` on the first block tspan when `options.y` is truthy, and the graph passes no `y`, so nothing absolute is set here. The `dy` was inert because every line tspan gets its own explicit `dy` (including `0em` for the first line), and per SVG text layout rules the innermost `dy` supersedes the ancestor `<text>` value for those glyphs.

Verified empirically in the dev app: toggling `dy="0.32em"` on a rendered label shifts the bbox by exactly 0px. So the removal is a visual no-op — good to keep.

🤖 Posted via [Claude Code](https://claude.com/claude-code)

rokotyan:
Thanks for the two fixes @lee00678! I reviewed both — notes below, plus a follow-up commit.

**cb60844d (remove `dy`)** — correct, keeping it. See the inline thread for a small correction on the mechanism (no absolute `y` is set here; the line tspans' own `dy` values supersede the `<text>`-level one).

**c8efadf1 (sub-label-only background)** — right bug and right idea; you were also right that it pre-dates this PR (`main` only ever checked `nodeLabel`). But adding a required `subLabel` parameter changed `setLabelRect` to 4 arguments while `modules/panel/index.ts` still calls it with 3, so the arguments shift: `panelSelectors.labelText` lands in `subLabel` and `selector` becomes `undefined`. `tsc` flags both call sites with `TS2554: Expected 4 arguments, but got 3` (builds stay green because ts-loader transpiles past type errors — same reason CI didn't catch this in #868), and at runtime hovering any panel label throws `TypeError: Cannot read properties of null (reading 'getBBox')` — reproduced in the dev app's `Layout: ELK` example.

**Follow-up: 14bd1965** — `setLabelRect(labelSelection, selector)` now derives emptiness from the rendered text element (`labelTextSelection.text()`) instead of taking label/sub-label strings, and the node + panel call sites are updated. That keeps your sub-label-only background fix working, removes the argument-ordering hazard for good, and handles the hover-expanded state automatically. Verified in the dev app: node label hover expand/restore works, the background rect shows for sub-label-only nodes, and panel label hover no longer throws.

🤖 Posted via [Claude Code](https://claude.com/claude-code)

lee00678 (approved):
Verified and the finally fix looks good. cmds-report passes as well.

PR #885, opened 2026-08-17
Component | Donut: A new `minSegmentAngle` config option

rokotyan:
Adds a new `minSegmentAngle` config option to `Donut`

## Problem

`Donut` applies `padAngle` at every segment boundary, and d3's arc generator insets both ends
of each arc by more than `padAngle / 2` (its default `padRadius` exceeds either radius). Any
segment with a small non-zero value whose proportional sweep is smaller than that inset
collapses to a zero-area path and silently disappears, exposing the donut background.

Reproduction: `data = [6800, 2400, 790, 10]` with `padAngle: 0.02` — the `10` segment doesn't
render. The existing `showEmptySegments` / `emptySegmentAngle` mechanism only catches
exact-zero values, so tiny non-zero segments fall through.

## Solution

`minSegmentAngle` (radians, default `undefined` = current behavior) guarantees a minimum
angular sweep for every segment with a positive value:

- Inflation happens in angle space after the pie layout — `datum.value`, events, tooltips,
  and legend items keep reporting the raw input values.
- The inflated angle is taken proportionally from the remaining segments; the set of inflated
  segments is settled iteratively, since inflating one segment can push the next-smallest
  below the threshold.
- If the combined minimum exceeds the angle range, the minimum is scaled down so the total
  still spans `angleRange` (works with half donuts).
- Zero/negative values keep the existing `showEmptySegments` handling unchanged.
- A sensible setting is `minSegmentAngle >= 2 * padAngle` (documented in the prop's JSDoc);
  the two options are not implicitly coupled.

- [x] Dev examples
- [x] Wrappers
- [ ] Gallery example (not applicable)
- [x] Docs (includes an interactive `minSegmentAngle` slider example)

SCR-20260817-ixbf

https://github.com/user-attachments/assets/cd67f2c9-b813-4317-9dff-5c0abdcf070c

### DIFF
--- a/packages/angular/src/components/donut/donut.component.ts
+++ b/packages/angular/src/components/donut/donut.component.ts
@@ -83,6 +83,14 @@ export class VisDonutComponent<Datum> implements DonutConfigInterface<Datum>, Af
   /** Pad angle. Default: `0` */
   @Input() padAngle?: number
 
+  /** Minimum angular sweep, in radians, for any segment with a positive value.
+   * Segments whose proportional sweep would be smaller are inflated to this angle,
+   * with the difference taken proportionally from the remaining segments.
+   * Guarantees a segment can survive the `padAngle` inset instead of collapsing
+   * to a zero-area path; a sensible setting is `minSegmentAngle >= 2 * padAngle`.
+   * Default: `undefined` */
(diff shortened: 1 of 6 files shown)


lee00678 (approved):
LGTM

PR #886, opened 2026-08-18
Misc | Vitest: Setup vitest with SSR support, adding it to the CI pipeline

lee00678:
This PR adds Vitest to Unovis.
Also added it as part of the pull-request CI pipeline.
It only adds one test, as adding tests are not part of the set up. I will add more in a follow up MR.

For adding tests, I propose the following:
ts (core) — full coverage. This is where all real logic and all the SSR risk (window/document access) actually lives. High value.

Wrapper packages (react/angular/svelte/vue/solid) — test the shape, not every component.

The per-component files (VisDonut, VisLine, etc.) are template output. Testing all of them is testing the template N times over — low value, high maintenance (they'd all need regenerating in sync if the template changes).
Instead, add a small representative sample per framework: one XY component, one standalone/core component, one composite (since those three shapes have genuinely different generated code paths) — mount, update props, unmount, and (for SSR) run through that framework's own SSR render path (renderToString for React/Solid, Angular Universal, Svelte SSR compile, Vue SSR) without throwing. That validates the template/generator for every component of that shape at once.
Test the genuinely hand-written code directly, which is where real bugs live:
react.ts (arePropsEqual)
each framework's composites/ (e.g. time-series) — not generated, has real logic
the wrapper generator itself in integrations — this is arguably the highest-leverage target, since a bug there breaks every generated file across every framework at once. A unit test on the generator's output is more valuable than N per-component tests downstream of it.
Skip: barrel files (components.ts), and per-component generated files beyond the representative sample.

## Checklist

- [x] Dev examples
- [x] Dev gallery examples (`pnpm dev:gallery`)
- [x] Gallery example
- [ ] Docs

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -14,6 +14,17 @@ jobs:
     - run: pnpm install --frozen-lockfile
     - run: pnpm run build && pnpm run build:dev
 
+  test:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-node@v3
+      with:
+        node-version-file: '.nvmrc'
+    - uses: pnpm/action-setup@v4
+    - run: pnpm install --frozen-lockfile
+    - run: pnpm run test:ts
+
   lint:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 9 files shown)


suryahanumandla:
LGTM!

suryahanumandla (approved):

PR #891, opened 2026-08-25
Clear vite-plugin-dts type errors and fail the build on missing entry declarations

### LINKED ISSUE #890: vite-plugin-dts silently skips dist/index.d.ts and dist/maps.d.ts due to pre-existing type errors
## Problem

Since the vite migration, the published `dist` has no root type entries: `vite-plugin-dts` skips emitting declarations for entries whose type-check reports diagnostics, and pre-existing diagnostics in the source stop `dist/index.d.ts` and `dist/maps.d.ts` from being generated — while all ~250 inner declaration files emit fine. The build exits 0, so nothing flags it.

Result: `@unovis/ts` resolves to no types at the package root (`./index.ts` → `'./dist'` → nothing). Downstream type-checks fail with errors like `Property 'Axis' does not exist on type 'typeof import("@unovis/ts")'`, and a published dist would ship untyped at the root.

#873 papers over it with a post-build step (`packages/ts/build-entry-dts.mjs`) that writes the two entry files from the barrels that do emit — that keeps consumers typed, but the real fix is clearing the diagnostics so the plugin emits them itself.

## The diagnostics

`npm run forcebuild` in `packages/ts` prints them; a sample of the ~30:

- `components/annotations/index.ts:28` — TS2612: Property 'g' will overwrite the base property (needs `declare`)
- `components/axis/index.ts:324` — TS7006/TS2345: implicit any in a tick filter
- `components/brush/index.ts:177` — TS2345: `number[]` vs `BrushSelection`
- `components/bullet-legend/index.ts:85`, `chord-diagram/index.ts:143,206`, and similar accessor-variance mismatches elsewhere

Most look mechanical (`declare` modifiers, explicit parameter types, a few casts at d3 boundaries).

## Suggested guardrail

Once clean, fail the build when the entry declarations are missing — the silent-skip behavior is what let this land unnoticed. Happy to take the cleanup as a PR series (it's ~30 sites, mostly one-liners).

rokotyan:
Closes #890

Since the vite migration, `vite-plugin-dts` type-checks the package on every build — something the old rollup setup deliberately skipped (`check: false`, with a TODO acknowledging it). The accumulated diagnostics (~30 across 20 files) can block the plugin from emitting the root entry declarations `dist/index.d.ts` / `dist/maps.d.ts`, and the build exited 0 regardless. This PR clears every diagnostic and makes the build fail loudly if the entries ever go missing again.

- [x] Type fixes across `packages/ts` (compile-time only — no emitted-JS behavior change)
- [x] Build guardrail: `afterBuild` entry check in `vite.config.ts` + fail-fast `&&` chaining in the build scripts

## Why each kind of change is safe

**1. `declare` modifiers on narrowed base-class fields** (`g`, `element`, `config`, `prevConfig`, `_svgDefs` in annotations, graph, leaflet-map, sankey, topojson-map, xy-component, xy-container)
`declare` is erased from the emitted JS entirely — the field only existed to narrow the base class's type, and the base constructor still performs the actual assignment. This is also the *correct* form under `useDefineForClassFields` semantics, where a bare re-declaration would clobber the base's value with `undefined` (the hazard TS2612/TS2729 warn about).

**2. Removed extra `.call()` arguments** (graph `createLinks`/`createPanels`/`zoomLinks`, chord `createNode`/`updateNode`, sankey `removeLinks`, timeline `_renderLines`)
d3's `selection.call(fn, ...args)` forwards the extra args to `fn` — but none of these callees declare parameters for them, and none reads `arguments` or rest params. JavaScript silently discards surplus arguments, so the values were dead on arrival; removing them is a runtime no-op. TS2554 was flagging the call sites, not the functions.

**3. Generic propagation instead of fixed concrete types** (chord `ChordRibbon<N, L>`/`getRibbons`, graph `updatePanels`/`removePanels`, topojson `calculateClusterIndex`/`geoJsonPointToScreenPoint`)
Types only — zero runtime code. The signatures now accept the caller's own generics instead of variance-broken concrete ones, so they *widen* what compiles; every existing valid call still type-checks (all call sites are internal and rely on inference — verified by grep across the monorepo).

**4. Precise d3 generics and boundary casts** (`partition<T>()`, `arc<T>()` anim-state types, `Selection$Transition` for timeline, tuple-typed brush range, `select<SVGSVGElement>`, plus a few one-line `as` casts in axis/tooltip/color/leaflet utils)
Type annotations and casts are erased at compile time. Each cast narrows to what the value already is at runtime (e.g. tooltip's `html` is provably a string after its guards; the legacy `key === true` path in `getColor` never reaches an accessor with a non-string).

**5. TopoJSON map: group creation moved from field initializers into the constructor**
The only real code motion. Field initializers already ran after `super()` (so `this.g` existed) and before the constructor body; they now run at the top of the constructor body in the *same relative order*, so the DOM `append` sequence — which controls paint layering — is unchanged. This matches the existing pattern in graph, chord-diagram, and sankey.

**6. Reordered chained setters** (plotband `.text()` after `.attr()`)
Independent setters on the same element; order between them has no observable effect. The reorder just keeps the chain on a type where `.attr` remains available.

## Guardrail

`vite-plugin-dts` logs diagnostics but doesn't fail the build, and `forcebuild` chained commands with `;` — so a build that skipped the entries still exited 0 (the "nothing flags it" part of #890). Now:
- an `afterBuild` hook throws if `dist/index.d.ts` or `dist/maps.d.ts` wasn't emitted (verified it fires by requiring a bogus entry — build exits 1);
- `forcebuild` chains with `&&` and `build` no longer writes `.srcsha` on failure (so a failed build can't be cached as good).

## Verification

- `npm run forcebuild`: **0 diagnostics**, 252 `.d.ts` files emitted including both entries, exit 0
- `pnpm build:react` and `pnpm build:dev` (CI's checks) pass against the rebuilt core
- ESLint clean on all changed files (remaining warnings confirmed pre-existing)

### DIFF
--- a/packages/ts/package.json
+++ b/packages/ts/package.json
@@ -47,8 +47,8 @@
     "dist/styles/index.js"
   ],
   "scripts": {
-    "build": "sha=$(tar cf - ./src | shasum); if [[ $(echo $sha) == $(< .srcsha) ]] && [[ -d \"./dist\" ]]; then echo \"Lib Build Exists\"; else npm run forcebuild; echo $sha > .srcsha; fi; npm run build:package-json",
-    "forcebuild": "rimraf dist; vite build; rm -rf dist/.cache; cp LICENSE README.md ./dist; npm run build:package-json",
+    "build": "sha=$(tar cf - ./src | shasum); if [[ $(echo $sha) == $(< .srcsha) ]] && [[ -d \"./dist\" ]]; then echo \"Lib Build Exists\"; else npm run forcebuild || exit 1; echo $sha > .srcsha; fi; npm run build:package-json",
+    "forcebuild": "rimraf dist && vite build && rm -rf dist/.cache && cp LICENSE README.md ./dist && npm run build:package-json",
     "publish:dist": "cd ./dist; npm publish",
     "build:package-json": "node -e \"const p=JSON.parse(require('fs').readFileSync('package.json','utf8')); delete p.typesVersions; require('fs').writeFileSync('./dist/package.json', JSON.stringify(p, null, 2))\"",
     "license:check": "pnpm license-report --config=../../lic-report-config.json > licences.md"
--- a/packages/ts/src/components/annotations/index.ts
+++ b/packages/ts/src/components/annotations/index.ts
@@ -25,7 +25,7 @@ export class Annotations extends ComponentCore<unknown[], AnnotationsConfigInter
   protected _defaultConfig = AnnotationsDefaultConfig as AnnotationsConfigInterface
   public config: AnnotationsConfigInterface = this._defaultConfig
 
-  g: Selection<SVGGElement, unknown, null, undefined>
+  declare g: Selection<SVGGElement, unknown, null, undefined>
 
   events = {}
 
--- a/packages/ts/src/components/axis/index.ts
+++ b/packages/ts/src/components/axis/index.ts
@@ -317,7 +317,7 @@ export class Axis<Datum> extends XYComponentCore<Datum, AxisConfigInterface<Datu
       .classed(s.tickLabelHideable, Boolean(config.tickTextHideOverlapping))
       // Ticks outside the fitted set render as unlabeled tick marks
       .classed(s.tickTextHidden, tickValue => labeledTickKeys ? !labeledTickKeys.has(tickKey(tickValue)) : false)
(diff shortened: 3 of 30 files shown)


lee00678:
pnpm cmds-report passes.
Forcebuild in packages/ts pass without any error.

lee00678 (approved):

PR #893, opened 2026-08-26
Core: Support CSP nonce via global

suryahanumandla:
## What & why
This MR introduces opt-in Content Security Policy (CSP) nonce support for Unovis. Because every Unovis component injects <style> elements into document.head via Emotion, those tags are rejected by a strict style-src CSP directive unless they carry a nonce. The change consists of three logical commits:

Core — A new @/styles/emotion wrapper module (packages/ts/src/styles/emotion.ts) creates a named Emotion instance using @emotion/css/create-instance, reading globalThis.UNOVIS_NONCE at module evaluation time and forwarding it as the Emotion cache's nonce. 

Shared / Gallery — A Vite plugin added to packages/shared/vite.config.ts enables an end-to-end CSP verification harness (activated by UNOVIS_CSP=1). It sets a real Content-Security-Policy response header, adds nonces to every <script>/<style> tag in the HTML transform, and injects a bootstrap script that stamps the nonce onto dynamically created elements via document.createElement and Node.prototype monkey-patching.

Website — A new csp.mdx guide covers the Quick Start, the recommended CSP header, per-framework integration snippets (React/Next.js, Angular, SvelteKit, Vue/Nuxt, SolidStart, vanilla TS), known limitations, and local verification instructions.

## Checklist

- [ ] Dev examples
- [ ] Dev gallery examples (`pnpm dev:gallery`)
- [ ] Wrappers (regenerated via `pnpm generate`)
- [ ] Gallery example
- [ ] Docs
- [ ] Lint passes on the files I changed
- [ ] Builds locally (`pnpm build`)
- [ ] All commands pass (`pnpm cmds-report`)

## Screenshots / recordings

image

### DIFF
--- a/ai/skills/add-component/SKILL.md
+++ b/ai/skills/add-component/SKILL.md
@@ -26,7 +26,9 @@ Create `packages/ts/src/components/<kebab-name>/` (kebab-case dir + files):
   and a `_render(customDuration?)` (or `render`) with clear **enter / update / exit** D3 selections.
 - **`style.ts`** — emotion `css` selectors + a `cssVarDefaults` map. CSS variables are named
   `--vis-<component>-<selector>-<property>`; every color var needs a `--vis-dark-...` counterpart.
-  Export `variables = getCssVarNames(cssVarDefaults)` and call `injectGlobalCssVariables(...)`.
+  Import `css` / `injectGlobal` from `@/styles/emotion` (not `@emotion/css` directly) so styles
+  respect the `globalThis.UNOVIS_NONCE` CSP nonce. Export `variables = getCssVarNames(cssVarDefaults)`
+  and call `injectGlobalCssVariables(...)`.
 - **`types.ts`** — component-specific types (optional).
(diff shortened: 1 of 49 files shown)


rokotyan:
This is great!

lee00678:
This looks good. Minor update in the cmds-report to remove gallery:csp from the script. 
When given a variable, I'm able to see it in the style tag. Headers also looks correct with nouce added.

lee00678:
Emotion doesn't allow the same prefix as `css-` when allowing nouce. But I realized this could be a breaking change if use have some tests setup with specific `css` class names to get the elements. 
So adding the fallback, if Unovis_nouce is enabled, prefix updates to `unovis-`, otherwise it stays as `css-`

`pnpm dev:gallery`
Screenshot 2026-08-28 at 11 37 46 AM

`pnpm dev:gallery:csp`
Screenshot 2026-08-28 at 11 37 20 AM

 cc @suryahanumandla

lee00678 (approved):

PR #704, opened 2026-01-20
Misc | Angular: Upgrade angular from 12 to 20 LTS (Do not merge until 1.7)

suryahanumandla:
-  Upgraded angular from 12 to 20
-  Fixed gallery issue
-  Fixed vulnerabilities 
-  Generated licenses

### Verified all commands and green
- [x] pnpm dev:gallery
- [x] pnpm dev
- [x] pnpm cmds-report
- [x] pnpm website
- [x] cd package/angular && pnpm gallery

# vulnerabilities 

Before:
before

After:
after

after-vul-report

PR #895, opened 2026-08-26
Angular | Fix: Package.json main, module, typings paths for ng-packagr 20 output

lee00678:
In this pr, the path was not updated. 1.6.7 (our latest version is not impacted)
When installing the latest patch, got path error for unovis/angular. 
Traced back to this:
https://github.com/f5/unovis/pull/704/changes#diff-d08139eef05218816a5597154199c170727c5bcd22546a9e6367735e6b223fb0R37-R39

### DIFF
--- a/packages/angular/gallery/tsconfig.app.json
+++ b/packages/angular/gallery/tsconfig.app.json
@@ -22,7 +22,7 @@
     "noUnusedLocals": true,
     "importHelpers": true,
     "typeRoots": [
-        "node_modules/@types"
+        "../node_modules/@types"
     ],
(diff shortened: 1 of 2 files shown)


suryahanumandla:
angular package gallery is failing, otherwise changes are looking good. Looks like, this is an existing bug.

suryahanumandla:
> angular package gallery is failing, otherwise changes are looking good. Looks like, this is an existing bug.

fixed it.

lee00678:
Tested by releasing a patch version of angular. Merging.

PR #896, opened 2026-08-26
StackedBar: Transition fixes and new `barStyle` accessor

rokotyan:
This PR:
- Fixes a StackedBar transition bug 
- Tweaks enter animation (bars now grow altogether from 0, instead of appearing in the center
- Adds a new config option for custom per-bar styling.

**Disappearing bars on rapid data updates.** The exit transitions' `.on('interrupt', remove)` handler (added in 98f8c9ad, shipped in 1.6.6) deletes bars that were legitimately re-adopted by a subsequent data join: when a re-render arrives mid exit fade and the segment's data is back, `smartTransition`'s `interrupt()` fires the old exit handler and removes the live node. Fixed by dropping the handler on `bars.exit()` — an interrupted exit always means the next join re-adopted the node (it either re-enters the update selection, which now restores `opacity`, or exits again and gets a fresh fade + remove). The group-level handler stays: exiting groups swap their class, so they can never be re-adopted.

**Enter animation.** Entering bars used to grow out of the start of their own stacked span, so non-first segments appeared to grow from the middle of the bar. They now collapse onto the stack's origin — the zero line, or the datum's `baseline` value when set — so the whole stack grows out of it. Works for positive and negative segments in both orientations.

**New `barStyle` accessor.** `(datum, stackIndex) => StyleDeclaration` applies custom inline styles per bar (typed via `csstype`, camelCase or kebab-case keys, `--custom-property` support). Keys the component manages itself (`fill`, `opacity`, `cursor`, `mask`) are merged into the render transitions as target values, so they win over defaults like the `color` accessor while staying animated; other keys are applied instantly via a new reusable `applyInlineStyles` util that also cleans up keys no longer returned. Useful for e.g. dashed "projected value" bars.

---

- [x] Dev examples (`Enter & Exit Transitions` stress-test: rapid updates, same-tick double renders, negative stacks; `Per-Bar Styles`: actual-vs-projected chart with a movable boundary demonstrating `barStyle` application and cleanup)
- [x] Wrappers
- [x] Docs (new "Custom Bar Styles" section with a live example; `barStyle` appears in the auto-generated props table)

### Fixes the issue with "hanging" bars
SCR-20260826-jlur

### New dev examples
SCR-20260826-oxdu

https://github.com/user-attachments/assets/cfd86b87-e5f5-494b-bfe8-6101c23bb05b

### Docs update
SCR-20260826-paiq

### DIFF
--- a/packages/ts/src/components/stacked-bar/index.ts
+++ b/packages/ts/src/components/stacked-bar/index.ts
@@ -4,9 +4,20 @@ import { min, max } from 'd3-array'
 import { XYComponentCore } from '@/core/xy-component'
 
 // Utils
-import { isNumber, isArray, isEmpty, clamp, getStackedExtentWithBaseline, getString, getNumber, getStackedData, getExtent } from '@/utils/data'
+import {
+  isNumber,
+  isArray,
+  isEmpty,
+  clamp,
+  getStackedExtentWithBaseline,
+  getString,
+  getNumber,
+  getValue,
+  getStackedData,
+  getExtent,
+} from '@/utils/data'
 import { roundedRectPath } from '@/utils/path'
-import { smartTransition } from '@/utils/d3'
+import { smartTransition, applyInlineStyles } from '@/utils/d3'
 import { getColor } from '@/utils/color'
 import { getPattern, getFillPatternValue, UNOVIS_PATTERN_INDEX_ATTR } from '@/utils/pattern'
 
@@ -15,12 +26,16 @@ import { ContinuousScale } from '@/types/scale'
 import { NumericAccessor } from '@/types/accessor'
 import { Spacing } from '@/types/spacing'
 import { Orientation } from '@/types/position'
+import { StyleDeclaration } from '@/types/style'
 
 // Local Types
 import { StackedBarDataRecord } from './types'
 
 // Config
 import { StackedBarDefaultConfig, StackedBarConfigInterface } from './config'
+
+// Constants
+import { MANAGED_BAR_STYLES } from './constants'
 
 // Styles
 import * as s from './style'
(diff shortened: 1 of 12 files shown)


packages/ts/src/components/stacked-bar/index.ts
```diff
@@ -177,23 +192,30 @@ export class StackedBar<Datum> extends XYComponentCore<Datum, StackedBarConfigIn
...
+
+    // No `interrupt` removal here: an interrupted exit means the next data join has re-adopted the node
+    // (it keeps the `bar` class), so it either re-enters the update selection or exits and fades again.
+    // Removing it on `interrupt` would delete bars that legitimately re-entered on a rapid re-render.
     smartTransition(bars.exit(), duration)
       .style('opacity', 0)
       .remove()
-      // `transition.remove()` only fires on `end`; if the transition is interrupted by a re-render,
-      // the node would linger in the DOM with opacity < 1 and could be picked up by the next data join.
-      .on('interrupt', function () { this.remove() })
```
lee00678, line 196: what about this: https://github.com/rokotyan/unovis/blob/d217c13c9f961ff51ac2a16f7df373f4139fd172/packages/ts/src/components/stacked-bar/index.ts#L137

lee00678:
https://github.com/user-attachments/assets/b7ba57a6-e4f7-4e5f-a65e-af25b8451b93

Entering-bar origin change in _getBarPath: entering bars now always collapse to config.baseline/0 instead of d.stacked[0]. That's right for a brand-new group where every segment enters together. But if a single stack segment re-enters into an already-rendered group (e.g. it was filtered out for being zero-height last render, per the barMinHeight1Px filter a few lines up), it will now animate up from the baseline through the rest of the stack instead of growing in place from its already-correct position (the old behavior). Worth double-checking this is intended/acceptable, since it's a plausible regression path not obviously covered by the new dev examples (those mostly swap whole groups, not individual segments within a stable group).

lee00678 (approved):
Good fix!

PR #897, opened 2026-08-27
Plotline & Plotband: Fit labels into the chart and support styled text blocks

rokotyan:
`Plotline` and `Plotband` labels used to get clipped by the container's clipping path whenever they were placed near the edge of the chart — e.g. a plotline at the top of the domain, or the plotband's default `top-left-outside` label position.

What's changed:

- Both components now report `bleed`, so the container reserves space for the label outside of the scale range (capped at a third of the range). When the bleed can't help — the cap is hit, or the label overflows across the component's axis where it doesn't follow the scale — the label is shifted into the chart area instead, so it always stays visible.
- Labels render through the Unovis text engine: alignment is expressed with the `TextAlign` / `VerticalAlign` enums instead of the browser-quirky `dominant-baseline`, and the label's bounding box is measured from the exact same wrapped text layout the render uses.
- `labelText` now also accepts `UnovisText | UnovisText[]` for styled multi-line labels (per-block font size, weight, color, margins, wrapping width); plain strings support `\n` line breaks. Block-level values take priority over `labelSize` / `labelColor`.
- Engine fix that this surfaced: a text block following a larger one used to overlap it, because the first-line advance was computed in the new block's `em` units. The advance now derives from the previous block's line height (equal-size blocks render pixel-identically), with the height estimation kept in lockstep.
- Assorted fixes along the way: labels and lines no longer animate on chart resize, `null` values hide the component instead of drawing at zero, and stale labels/bands are cleared when their config is removed.

- [x] Dev examples (label bleed playgrounds covering every position/orientation/axis combination)
- [x] Wrappers

https://github.com/user-attachments/assets/41102143-9b52-4cac-8503-8446b11f191f

https://github.com/user-attachments/assets/da6ca5cb-7c25-4d48-ad91-b966878c2cfa


rokotyan:
cc @50rayn

50rayn:
Looks cool.
Is similar to #777 somehow, just with small differences:

- this one: label vs. container boundary — bleed + shift so labels aren't clipped by the chart edge
- #777: label vs. other plotline/plotband labels — auto-positioning when two labels would overlap each other

They're complementary, not duplicates.

lee00678 (approved):
Looks good.

PR #899, opened 2026-08-28
GroupedBar: Exit transition fix and new `barStyle` accessor

rokotyan:
Ports the applicable parts of #896 to _GroupedBar_:

**Fixes lingering groups on rapid data updates.** GroupedBar's group exit transition had no `interrupt` handler: when a re-render arrived mid exit fade, the group was never removed — its class had already been swapped to `barGroupExit`, so no later data join could re-adopt or clean it up, leaving a semi-transparent ghost group (with its bars frozen mid-squash) in the DOM forever. Fixed by removing the node on `interrupt`, mirroring StackedBar's group-level handling.

**New `barStyle` accessor.** Same feature as StackedBar's in #896: `(datum, indexWithinGroup) => StyleDeclaration` applies custom inline styles per bar. Keys the component manages itself (`fill`, `cursor`, `mask`) are merged into the render transitions as target values, so they win over defaults like the `color` accessor while staying animated; other keys are applied instantly and clean up when no longer returned. Unlike StackedBar, `opacity` is not in the managed set — GroupedBar never animates bar-level opacity.

> **Depends on #896** — the two `Core` commits (`applyInlineStyles`, `StyleDeclaration` export) are cherry-picked from it and will dedupe once it merges.

- [x] Dev examples (`Per-Bar Styles`: actual-vs-projected chart with a movable boundary demonstrating `barStyle` application and cleanup)
- [x] Wrappers (Angular regenerated via autogen; others regenerate to no diff)
- [x] Docs (new "Custom Bar Styles" section with a live example; `barStyle` appears in the auto-generated props table)

SCR-20260828-micf

SCR-20260828-miwo

### DIFF
--- a/packages/angular/src/components/grouped-bar/grouped-bar.component.ts
+++ b/packages/angular/src/components/grouped-bar/grouped-bar.component.ts
@@ -11,6 +11,7 @@ import {
   ContinuousScale,
   GenericAccessor,
   FillPatternType,
+  StyleDeclaration,
   StringAccessor,
   Orientation,
 } from '@unovis/ts'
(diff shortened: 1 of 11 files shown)


lee00678 (approved):
LG

PR #922, opened 2026-09-15
LeafletMap | MapLibre: Ship a bundler-independent worker

### LINKED ISSUE #921: LeafletMap | MapLibre: worker fails silently without consumer bundler config
## Problem

With maplibre-gl 6, `LeafletMap`'s MapLibre renderer fails silently in apps without maplibre-specific bundler config: the style background renders, no tiles, no error. The current `?maplibreWorkerAsset` workaround in `map.ts` only works if the consumer's bundler has a matching rule (our dev gallery carries it in `webpack.config.js`). f5/unovis#912 has the same trap.

Image

## Root cause

maplibre 6 no longer inlines its worker. `maplibre-gl.mjs` resolves it at runtime via `new URL('./maplibre-gl-worker.mjs', import.meta.url)`, and the worker imports `./maplibre-gl-shared.mjs` by relative path. Bundlers can't see this as a worker entry, so:

- Vite copies the `.mjs` verbatim; stock nginx serves it as `application/octet-stream` and the browser rejects the module worker. Fixing the MIME type isn't enough, since the sibling is emitted content-hashed and the relative import 404s into the SPA fallback.
- `vite dev` pre-bundling moves `import.meta.url` into `.vite/deps`, where the file doesn't exist.
- webpack ignores the dynamic `new URL()` entirely.

There's no UMD or inlined build in 6.x to fall back to.

## Who bundles the worker?

**A. The consumer** (status quo, f5/unovis#912): each app emits both `.mjs` files unhashed and adjacent (or uses `?worker&url`) and calls `setWorkerUrl()`.

- Pro: no maplibre artefact in `@unovis/ts`; worker and main thread always come from the same installed version.
- Con: every consumer repeats bundler-specific config, and forgetting it fails silently. Vite's default output is the broken one, so docs alone can't fix it.

**B. The library**: a Vite plugin in `packages/ts` bundles the installed worker into one self-contained script, emitted in `dist` as a module exporting the source string. `LeafletMap` starts it via a `Blob` URL unless `getWorkerUrl()` is already set.

- Pro: zero consumer config. Verified with `vite build` behind stock nginx, `vite dev`, and the webpack dev gallery. Build asserts the worker registration survived bundling, so an empty tree-shaken worker can't ship. The marker and dev webpack rules go away.
- Con: ~480 KB (~140 KB gzip) in a lazily loaded chunk; consumers need `worker-src 'self' blob:`; the worker is frozen at publish time, so `maplibre-gl` should stay pinned exactly and a runtime warning should flag a version mismatch.

Rejected: Vite `?worker&url` (same failure), `?worker&inline` (exports a factory, not a URL), committing the built worker, postinstall builds (skipped by pnpm 10 and `--ignore-scripts`).

## Proposal

Option B, with an exact `maplibre-gl` pin, a version-mismatch warning, and a CSP doc note. Keep the `getWorkerUrl()` guard so A stays possible. Long term, the fix belongs upstream: maplibre should create the worker in one statically analysable `new Worker(new URL(...), { type: 'module' })` expression, or ship an inlined build again. PR follows.

rokotyan:
Fixes #921

With maplibre-gl 6, `LeafletMap`'s MapLibre renderer silently renders no tiles in apps without maplibre-specific bundler config. This makes the published `@unovis/ts` work with zero consumer configuration, and removes the `?maplibreWorkerAsset` marker and the dev-gallery webpack rules that supported it.

- [x] Core
- [x] Dev examples (webpack config only, no new examples)
- [x] Docs (CSP guide)
- [ ] Wrappers (not affected)
- [ ] Gallery example (not affected)

## What changed

- `packages/ts/vite-plugin-maplibre-worker-source.ts` serves a virtual module: the installed `maplibre-gl/dist/maplibre-gl-worker.mjs` bundled into one self-contained IIFE (via Vite's `build()` API, the worker file as entry) and exported as a string, plus the maplibre version it was built from. It's emitted as `dist/components/leaflet-map/modules/maplibre-worker-source.js`, so consumers get an ordinary ES module.
- `map.ts` imports it lazily in the MapLibre branch and calls `setWorkerUrl(URL.createObjectURL(new Blob([source], { type: 'text/javascript' })))`. The `if (!maplibre.getWorkerUrl())` guard is kept, so a consumer's own `setWorkerUrl()` still wins. The Blob URL is created once and never revoked: maplibre re-reads the URL for every worker in its pool and again when the pool is rebuilt.
- The plugin fails the build if the worker's top-level `self.worker = new Worker(self)` registration is missing, if the chunk still imports anything, or if it uses `import.meta`. maplibre marks its dist as side-effect free, so a tree-shaken empty worker otherwise builds "successfully".
- `LeafletMap` warns when `maplibre.getVersion()` differs from the version the worker was built from, so a main-thread/worker mismatch is no longer silent.
- The dev gallery compiles `packages/ts/src` with webpack, which can't resolve a Vite virtual module and treats `virtual:` as a URL scheme (so `resolve.alias` doesn't apply). A `NormalModuleReplacementPlugin` maps the id to the built file in `packages/ts/dist`. The `maplibre-gl.mjs` parser rule stays: without it webpack's "Critical dependency" warning returns.

## Why this design

Alternatives and why they were rejected are in #921. In short: Vite's `?worker&url` has the same `import.meta.url` failure modes; `?worker&inline` exports only a Worker factory and revokes its Blob URL on start, while maplibre's only hook is `setWorkerUrl(string)`; a postinstall build is skipped by pnpm 10 and `--ignore-scripts`; a CDN worker URL adds a third-party runtime dependency.

## CSP

Consumers rendering the MapLibre renderer need `worker-src 'self' blob:`. maplibre itself already uses Blob workers for cross-origin URLs. The CSP guide now says so. (Side note: the guide previously claimed the ELK graph spawns a Blob worker; `elk.bundled.js` runs its layout in-thread via a fake worker, so that directive was never needed there. Happy to correct that in this PR or separately.)

## Verification

Throwaway Vite app consuming the packed `dist` via a `file:` tarball, with no maplibre-specific config:

| Scenario | Result |
| --- | --- |
| `vite build`, served by a static server mimicking deployed nginx (`default_type application/octet-stream`, stock `mime.types`, SPA fallback) | Tiles render. Worker URL is `blob:`. Only `.js` assets requested; no `.mjs`, no `maplibre-gl-shared*`, no failed requests, no console errors. |
| `vite dev` with default dependency pre-bundling | Tiles render. Worker module served from `node_modules/.vite/deps/`. |
| `maplibregl.setWorkerUrl('/custom-worker.js')` before the map mounts | Respected: that file is fetched, the bundled worker module is never loaded, tiles render. |
| Dev gallery (`pnpm dev`, webpack) Vector Map example | Tiles render, compiles with zero warnings. |

Also: `pnpm build` (core), `pnpm build:dev`, lint and type-check clean for the touched files. A second `pnpm build` is skipped by the `.srcsha` cache, whose key now includes `package.json`.

## Consumer cleanup this unblocks

Apps that worked around this can delete: their `?worker&url` bundle of `maplibre-gl-worker.mjs` and the `setWorkerUrl()` call that used it, a direct `maplibre-gl` dependency added only for that, and any Vite `external`/`entryFileNames` tweaks that kept the worker unhashed. Bundler asset rules emitting `maplibre-gl-{worker,shared}.mjs` can go too.

## Open question for maintainers

`maplibre-gl` is declared as `^6.7.0` here. Since the worker is frozen at publish time while the main-thread maplibre-gl resolves at the consumer's install time, an exact pin would make the two match by construction (package managers nest a matching copy for us). The runtime warning covers the mismatch either way; pinning exactly is a one-line change if you prefer it.


rokotyan:
@lee00678 I've not tested it properly yet on my end. The MapLibre update was kind of a breaking change so it would be nice to fix and publish as 1.7.1

lee00678:
I was just about to ask if you've tested this on your end. Because we don't have a good source of verification other than what we use here and the dev examples we have.

lee00678:
@rokotyan I think this looks good, we plan on release 1.7.1 tmw. Flag me if you want us to hold off.

suryahanumandla:
@rokotyan LGTM! 

There is a follow up PR required to cover if someone doesn't want to make use of blob(strict CSP), they can do it with workers. I'll be making the changes for it before 1.7.1 release.

suryahanumandla (approved):

PR #927, opened 2026-09-21
Axis: Fit and centre rotated wrapped tick labels; `tickTextOverlapTolerance` and `tickTextMaxLines` options

rokotyan:
Fixes the adaptive tick fitting for rotated, wrapped X axis labels, and adds two related tick label options.

### Rotated labels (fix)
A categorical chart with 10 long labels at `tickTextAngle: -90` and `tickTextAdaptiveSets` labeled only 5 of them, off-centre. Two causes, both from treating every label as horizontal:

- The wrap budget was the tick's horizontal slot. Rotated to -90°, a label's on-screen width is its line count times the line height, so denser sets shrank the wrap width, added lines and made every label wider — the fitting rejected the very set it was making room for. The budget now follows the angle and stops shrinking with the label count as the text turns across the axis. Unchanged at 0°.
- A wrapped block was anchored on its first line, sliding a rotated block sideways by half its extra height. The block is now centred on the tick, with the predicted geometry matching the render.

Rotated labels are compared in their own frame in both the fitting and the overlap pass, instead of via screen bounding boxes that overlap long before steep labels do (`hideOverlappingLabels` gains a `rotationAngle` option).

### New options
- `tickTextOverlapTolerance` — the minimum gap between tick labels in pixels (`0` by default). Replaces a hard-coded constant. Also fixes `rectIntersect` moving one edge by twice the tolerance, so a tolerance now means twice its gap rather than three times.
- `tickTextMaxLines` — caps the lines a wrapped label can take; longer labels are trimmed to fit according to `tickTextTrimType` (start, middle or end). Backed by new `maxLines` / `trimMode` text options in `getWrappedText`.

https://github.com/user-attachments/assets/c119ded4-6bc9-4d64-93e8-1e51d61ee6a5

https://github.com/user-attachments/assets/2f15a670-be13-4696-99dd-a06530bf6e0f

- [x] Dev examples (`Rotated Category Labels`, `Tick Label Max Lines`)
- [x] Wrappers (Angular regenerated)
- [ ] Gallery example (not applicable)
- [x] Docs (`tickTextOverlapTolerance`, `tickTextMaxLines`)
- [x] Specs (`axis.spec.ts`, `text.spec.ts`)


packages/ts/src/components/axis/index.ts
```diff
@@ -404,18 +400,43 @@ export class Axis<Datum> extends XYComponentCore<Datum, AxisConfigInterface<Datu
...
+    // Fair share of the axis: the slot a labeled tick owns
+    const slotWidth = this._containerWidth / (labelCount + 1)
+    if (!config.tickTextAngle) return slotWidth
+
+    // Neighbouring rotated labels keep apart either along the axis or across their lines. Once a
+    // line clears the next tick across, the text is bounded only by the margin depth (a third of the
+    // container height) projected onto it — wrapping tighter would just stack more lines across.
+    // Shallower labels can only keep apart along the axis, so their text is bounded by the slot,
+    // less the projection of a line's height on it
+    const angleRad = config.tickTextAngle / 180 * Math.PI
```
devgru: I'd recommend extracting this portion of the method to a separate fn

